PR #20, opened 2020-02-12
Feature/msid

zakajd:
Added MSID wrapper for code from this repo: github.com/xgfs/msid 

I tried to rewrite it using only PyTorch, but there are some problems with PyTorch sparse tensors, so I remained everything in numpy and just cast tensors to match that type.


photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+        channel_reduction: Specifies the reduction to apply to the output:
+            ``'none'`` | ``'mean'`` | ``'sum'``. ``'none'``: no reduction will be applied,
+            ``'mean'``: the sum of the output will be divided by the number of
+            elements in the output, ``'sum'``: the output will be summed. 
+
+    """
+
+    def __init__(self, ts: List[float] = np.logspace(-1, 1, 256), k: int = 5, m: int = 10, niters: int = 100, rademacher: bool = False, graph_builder: str = 'sparse',
+              normalized_laplacian: bool = True, normalize: str = 'empty', msid_reduction: str = "max", channel_reduction: str = 'mean'):
+
```
snk4tr: Formatting: remove blank line here.

From the Google style guide (section 3.5):
Two blank lines between top-level definitions, be they function or class definitions. One blank line between method definitions and between the class line and the first method. **No blank line following a def line**. Use single blank lines as you judge appropriate within functions or methods.

photosynthesis_metrics/tv.py
```diff
@@ -0,0 +1,153 @@
+r"""Implemetation of Total Variation metric, based on article
```
snk4tr: This file seems like the exact copy of the file with this name from the previous pull request. Did you include this on purpose? If so, what is your motivation?

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+                                                    msid_mode=msid_reduction,
+                                                    normalized_laplacian=normalized_laplacian, 
+                                                    normalize=normalize)
+
+
+        # Generic loss parameters.
+        self.channel_reduction = channel_reduction
+
+        # Loss-specific parameters.
+
```
snk4tr: It is not clear what these lines are for

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+        N, _, _, _ = x.shape
+        result = []
+        for i in range(N):
+            x_i = x[i]
+            y_i = y[i]
+            
+            channels_x = [t.numpy() for t in map(torch.Tensor.squeeze_, torch.split(x_i, 1, dim=0))]
+            channels_y = [t.numpy() for t in map(torch.Tensor.squeeze_, torch.split(y_i, 1, dim=0))]
+
+            scores = torch.Tensor([self.msid_score(x, y) for x, y in zip(channels_x, channels_y)])
```
snk4tr: Please read MSID paper carefully and find that graphs should be built on embeddings

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+
+import torch
+import torch.nn.functional as f
+import torch.nn as nn
+import numpy as np
+
+import scipy.sparse as sps
+from scipy.sparse import lil_matrix, diags, eye
+
+from .utils import _validate_input
```
snk4tr: Consider removing unused imports

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+EPSILON = 1e-6
+NORMALIZATION = 1e6
+
+# ---- laplacian.py
+def np_euc_cdist(data):
```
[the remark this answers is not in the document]
zakajd: No need to be public. Changed this

photosynthesis_metrics/tv.py
zakajd: Discussed during weekly meeting.

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,420 @@
...
+        dists = dd - 2*data[i, :].dot(data.T)
+        inds = np.argpartition(dists, k+1)[:k+1]
+        inds = inds[inds!=i]
+        spmat[i, inds] = 1
+            
+    return spmat.tocsr()
+
+
+def construct_graph_kgraph(data, k):
+    import pykgraph
```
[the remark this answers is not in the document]
zakajd: Moved import on top, but there is a problem with that library.
It should be [compiled from source ](https://github.com/aaalgo/kgraph#building-and-installation) and only used in `_construct_graph_kgraph` function. There is other function `_construct_graph_sparse` which is being called by default and produces more accurate result (according to the paper). This library may be needed for really big datasets (like Imagenet), but its unclear for now how faster `kgraph` method comparing to `sparse`. 
For now I commented this line, because it's unclear if we should have such strong dependancy for just one line of code.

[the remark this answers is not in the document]
snk4tr: Regarding the concerns, I would propose to get rid of this lib and the function that uses it util the moment when we actually need it. For now it seems like unnecessary complication.

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,406 @@
+r"""Implemetation of Multi-scale Evaluation metric, based on paper
+ https://arxiv.org/abs/1905.11141 and author's repository https://github.com/xgfs/msid
+"""
+from functools import partial
+from typing import List, Tuple, Optional, Callable
```
snk4tr: Remove unused imports
```suggestion
from typing import List, Tuple
```

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,406 @@
...
+
+import torch
+import torch.nn as nn
```
snk4tr: Remove unused import

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,406 @@
...
+import numpy as np
+
+import scipy.sparse as sps
```
snk4tr: Remove unused import

photosynthesis_metrics/msid.py
```diff
@@ -0,0 +1,406 @@
...
+from scipy.sparse import lil_matrix, diags, eye
+
+from photosynthesis_metrics.feature_extractors.fid_inception import InceptionV3
```
snk4tr: Remove unused import

PR #31, opened 2020-04-24
Refactoring/issue#30

### LINKED ISSUE #30: Unify interface for feature metrics, remove redundant code
Right now for all metrics, that require features, there is exist a function to compute it AND a class, which both have almost the same functionality. The only difference is that inside a class there is also an option to precompute features from an image. 

Idea is to move everything related to metric computation into `compute_metric` class method, 
make a unified `forward` call which takes torch.Tensors and return torch.Tensor, moves them from GPU device to CPU and detaches if needed. 

Alternatively, we could add an option to automatically turn images into features inside a class forward call, but as discussed with @snk4tr, initially proposed solution is better.

zakajd:
Closes #30 
@snk4tr look and merge, if everything seems OK

### DIFF
--- a/photosynthesis_metrics/__init__.py
+++ b/photosynthesis_metrics/__init__.py
@@ -1,5 +1,5 @@
-from .ssim import ssim, multi_scale_ssim, SSIMLoss, MultiScaleSSIMLoss
-from .msid import MSID, compute_msid
-from .fid import FID, compute_fid
-from .kid import KID, compute_polynomial_mmd
-from .tv import total_variation, TVLoss
+from .ssim import SSIMLoss, MultiScaleSSIMLoss
+from .msid import MSID
+from .fid import FID
+from .kid import KID
+from .tv import TVLoss, total_variation
--- a/photosynthesis_metrics/__init__.py
+++ b/photosynthesis_metrics/__init__.py
@@ -1,4 +1,4 @@
-from .ssim import SSIMLoss, MultiScaleSSIMLoss
+from .ssim import ssim, multi_scale_ssim, SSIMLoss, MultiScaleSSIMLoss
 from .msid import MSID
 from .fid import FID
 from .kid import KID
--- a/photosynthesis_metrics/base.py
+++ b/photosynthesis_metrics/base.py
@@ -59,5 +59,6 @@ class BaseFeatureMetric(torch.nn.Module):
 
         return torch.cat(total_feats, dim=0)
 
-    def compute_metric(self, predicted_features: torch.Tensor, target_features: torch.Tensor):
+    @staticmethod
+    def compute_metric(predicted_features: torch.Tensor, target_features: torch.Tensor):
         raise NotImplementedError("This function should be defined for each children class")
--- a/photosynthesis_metrics/base.py
+++ b/photosynthesis_metrics/base.py
@@ -19,6 +19,7 @@ class BaseFeatureMetric(torch.nn.Module):
     def forward(self, predicted_features: torch.Tensor, target_features: torch.Tensor) -> torch.Tensor:
         # Sanity check for input
         _validate_features(predicted_features, target_features)
+        return self.compute_metric(predicted_features, target_features)
 
     def _compute_feats(
         self,
@@ -57,3 +58,6 @@ class BaseFeatureMetric(torch.nn.Module):
             total_feats.append(features[0].view(N, -1))
 
         return torch.cat(total_feats, dim=0)
+
+    def compute_metric(self, predicted_features: torch.Tensor, target_features: torch.Tensor):
+        raise NotImplementedError("This function should be defined for each children class")
--- a/photosynthesis_metrics/fid.py
+++ b/photosynthesis_metrics/fid.py
@@ -24,7 +24,7 @@ def __compute_fid(mu1: np.ndarray, sigma1: np.ndarray, mu2: np.ndarray, sigma2: 
     The Frechet Inception Distance between two multivariate Gaussians X_predicted ~ N(mu_1, sigm_1)
     and X_target ~ N(mu_2, sigm_2) is
         d^2 = ||mu_1 - mu_2||^2 + Tr(sigm_1 + sigm_2 - 2*sqrt(sigm_1*sigm_2)).
-    
+
     Args:
         mu1: mean of activations calculated on predicted samples
         sigma1: covariance matrix over activations calculated on predicted samples
@@ -93,12 +93,15 @@ class FID(BaseFeatureMetric):
         Then computes FID as d^2 = ||mu_1 - mu_2||^2 + Tr(sigm_1 + sigm_2 - 2*sqrt(sigm_1*sigm_2)).
 
         Args:
-            predicted_features: Samples from data distribution. Shape (N_samples, data_dim), dtype: torch.float32 in range 0 - 1.
-            target_features: Samples from data distribution. Shape (N_samples, data_dim), dtype: torch.float32 in range 0 - 1
+            predicted_features: Samples from data distribution.
+                Shape (N_samples, data_dim), dtype: torch.float32 in range 0 - 1.
+            target_features: Samples from data distribution.
+                Shape (N_samples, data_dim), dtype: torch.float32 in range 0 - 1
 
         Returns:
         --   : The Frechet Distance.
         """
+        # GPU -> CPU -> Numpy
         m_pred, s_pred = _compute_statistics(predicted_features.detach().cpu().numpy())
         m_targ, s_targ = _compute_statistics(target_features.detach().cpu().numpy())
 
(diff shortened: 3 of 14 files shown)


photosynthesis_metrics/fid.py
```diff
@@ -108,6 +85,26 @@ def forward(self, predicted_features: torch.Tensor, target_features: torch.Tenso
         Returns:
             score: Scalar value of the distance between image sets features.
         """
-        # Check inputs
-        super(FID, self).forward(predicted_features, target_features)
-        return self.compute(predicted_features, target_features)
+        super().forward(predicted_features, target_features)
```
denproc, line 88: Is it redundant as the method use the functionality of parent class? The doc string could be moved to the class description. The same applies for `KID` class @ kid.py and `MSID` @ msid.py

photosynthesis_metrics/ssim.py
```diff
@@ -159,7 +114,9 @@ def __init__(self, kernel_size: int = 11, kernel_sigma: float = 1.5, k1: float =
         # Cash kernel between calls.
         self.kernel = _fspecial_gauss_1d(kernel_size, kernel_sigma)
 
-    def forward(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    def forward(self,
```
denproc, line 117: Should there be the similar approach to interface for `_Loss` as it was introduced for `BaseFeatureMetric`? the same applies for other children of `_Loss`.

photosynthesis_metrics/__init__.py
```diff
@@ -1,5 +1,5 @@
-from .ssim import ssim, multi_scale_ssim, SSIMLoss, MultiScaleSSIMLoss
-from .msid import MSID, compute_msid
-from .fid import FID, compute_fid
-from .kid import KID, compute_polynomial_mmd
-from .tv import total_variation, TVLoss
+from .ssim import SSIMLoss, MultiScaleSSIMLoss
```
snk4tr, line 1: I do not like this part at all. It seems like we have only losses but not metrics which is not true

photosynthesis_metrics/fid.py
snk4tr: Agree

photosynthesis_metrics/kid.py
```diff
@@ -1,80 +1,19 @@
 from typing import Optional, Tuple, Union
 from functools import partial
```
snk4tr, line 2: Remove unused import

photosynthesis_metrics/base.py
```diff
@@ -19,6 +19,7 @@ def __init__(self) -> None:
     def forward(self, predicted_features: torch.Tensor, target_features: torch.Tensor) -> torch.Tensor:
         # Sanity check for input
         _validate_features(predicted_features, target_features)
+        return self.compute_metric(predicted_features, target_features)
 
     def _compute_feats(
```
snk4tr, line 24: Make this methods static since it does not use any class attributes

photosynthesis_metrics/utils.py
```diff
@@ -31,10 +30,9 @@ def _validate_input(x: torch.Tensor, y: torch.Tensor, kernel_size: Optional[int]
     if kernel_size is not None:
         assert kernel_size % 2 == 1, f'Kernel size must be odd, got {kernel_size}.'
     if scale_weights is not None:
-        assert isinstance(scale_weights, (list, tuple)), \
+        assert isinstance(scale_weights, (list, tuple, torch.Tensor)), \
             f'Scale weights must be of type list or tuple, got {type(scale_weights)}.'
-        assert len(scale_weights) == 4, f'Scale weights collection must contain 4 values, got {len(scale_weights)}.'
-    
+        assert len(scale_weights) == 5, f'Scale weights collection must contain 5 values, got {len(scale_weights)}.'
```
snk4tr, line 35: Could you comment on why it changed from 4 to 5 during the refactoring?

photosynthesis_metrics/utils.py
```diff
@@ -53,4 +51,6 @@ def _validate_features(x: torch.Tensor, y: torch.Tensor, ) -> None:
         f"Target features must have shape  (N_samples, encoder_dim), got {y.shape}"
     assert x.shape[1] == y.shape[1], \
         f"Features dimensionalities should match, otherwise it won't be possible to correctly compute statistics. \
-            Got {x.shape[1]} and {y.shape[1]}"
\ No newline at end of file
+            Got {x.shape[1]} and {y.shape[1]}"
+    assert x.device == y.device, \
+        f"Both tensors should be on the same device"
```
snk4tr, line 56: Like it!

photosynthesis_metrics/fid.py
zakajd: I thought about moving this to parent class. 
Those metrics differ in shapes of feature vectors they can work on. KID needs the same sample size for both input and target features. MSID and FID can work even for different number of samples. It's important, because for those metrics we can compute target features on train and compare them with input features from validation, which  is usually much smaller.

photosynthesis_metrics/utils.py
zakajd: Because `len((0.0448, 0.2856, 0.3001, 0.2363, 0.1333)) == 5`, it's weights from paper.  The reason this tests never failed before is that they were never called (scale_weights were equal to None)

photosynthesis_metrics/fid.py
denproc: Could not agree with you more, the differences in processing are taken into account @ `compute_metric`. Therefore, the only new information/difference at `forward` is doc string.

photosynthesis_metrics/ssim.py
snk4tr: I consider this extra level of abstraction to be redundant

zakajd:
@snk4tr  @denproc 
Check this PR and let's merge it, because all other SSIM and MS-SSIM related issues are inherently depend on it.

Pull request from @denproc  with matrix square root is already merged here with small adjustments, to allow faster computation on GPU (see comparison below).
image

tests/test_ssim.py
```diff
@@ -78,6 +78,36 @@ def test_ssim_raises_if_wrong_kernel_size_is_passed(prediction: torch.Tensor, ta
 
 
 # ================== Test class: `SSIMLoss` ==================
+def test_SSIM_measure_is_zero_for_equal_tensors(target: torch.Tensor) -> None:
```
denproc, line 81: Duplicates exiting code

tests/test_ssim.py
```diff
@@ -78,6 +78,22 @@ def test_ssim_raises_if_wrong_kernel_size_is_passed(prediction: torch.Tensor, ta
 
 
 # ================== Test class: `SSIMLoss` ==================
+def test_SSIM_measure_is_less_or_equal_to_one() -> None:
```
denproc, line 81: Duplicates `test_ssim_loss_is_less_or_equal_to_one` (cuda as well). Sorry for not specifying explicitly, thought you would check for all duplicates. This is the last one.

tests/test_ssim.py
```diff
@@ -187,6 +203,7 @@ def test_multi_scale_ssim_raises_if_wrong_kernel_size_is_passed(prediction: torc
 
 
 # ================== Test class: `MultiScaleSSIMLoss` ==================
+
```
denproc, line 206: Minor one, in other sections there are no empty line.

snk4tr:
> @snk4tr @denproc
> Check this PR and let's merge it, because all other SSIM and MS-SSIM related issues are inherently depend on it.
> 
> Pull request from @denproc with matrix square root is already merged here with small adjustments, to allow faster computation on GPU (see comparison below).
> image

Nice! This one definitely need to be added in docs as out library's advantage!

photosynthesis_metrics/kid.py
```diff
@@ -212,49 +148,117 @@ def _sqn(tensor : torch.Tensor) -> torch.Tensor:
...
-        coef0: Kernel parameter. See paper for details
-        var_at_m: Kernel variance. Default is `None`
-        ret_var: Whether to return variance after the distance is computed.
-                    This function will return Tuple[float, float] in this case. Default: False
+        predicted_features: Low-dimension representation of predicted image set. Shape (N_pred, encoder_dim)
+        target_features: Low-dimension representation of target image set. Shape (N_targ, encoder_dim)
+
+    Returns:
+        score: Scalar value of the distance between image sets features.
+        variance (optional): If `ret_var` is True, also returns variance
```
snk4tr, line 162: Specify types here since there are no type annotations:
```suggestion
        predicted_features(torch.Tensor): Low-dimension representation of predicted image set. Shape (N_pred, encoder_dim)
        target_features(torch.Tensor): Low-dimension representation of target image set. Shape (N_targ, encoder_dim)

    Returns:
        score(torch.Tensor): Scalar value of the distance between image sets features.
        variance(Optional[torch.Tensor]): If `ret_var` is True, also returns variance
```

PR #41, opened 2020-05-12
feature: Geometry Score

### LINKED ISSUE #19: Add Geometry Score (GS)
It's used in several articles to compare GAN quality. Authors of MSID score states, that it's better than GS.

https://github.com/KhrulkovV/geometry-score
https://arxiv.org/pdf/1802.02664.pdf

zakajd:
Closes #19.
I took official implementation repo and added faster computation. 
By default it worked only on 1 CPU, which was super slow for big number of iterations. Now number of processes can be set, which gives 4-8x boost.


photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+        L: Array of size l x d, containing landmarks
+
+    Returns:
+        distances: 3D array of size w x l x 2. It satisfies the property that
+            distances[i, :, :] is [idx_i, dists_i], where dists_i are the sorted distances
+            from the i-th witness to each point in L and idx_i are the indices of the corresponding points
+            in L, e.g., D[i, :, :] = [[0, 0.1], [1, 0.2], [3, 0.3], [2, 0.4]]
+        max_dist: Maximal distance between W and L
+    """
+    a = cdist(W, L)
```
denproc, line 84: Even though there is only lumpy implementation, we still could minimise `scipy` usage. `torch.cdist` could be the option, but it requires tensors as an arguments.

tests/test_gs.py
```diff
@@ -0,0 +1,63 @@
...
+        f'For same distributions GS should be small, got {diff}'
+
+
+@pytest.mark.skipif('gudhi' not in sys.modules, reason="Requires gudhi library")
+def test_GS_differs_for_not_simular_distributions(
+        features_prediction_beta: torch.Tensor, features_target_normal: torch.Tensor) -> None:
+    metric = GS(sample_size=64, num_iters=500, i_max=100, num_workers=4)
+    diff = metric(features_prediction_beta, features_target_normal)
+    assert diff >= 1.0, \
+        f'For different distributions GS diff should be big, got {diff}'
```
denproc, line 63: Should we add tests to check input types, dimensionality, device and output correctness compared to original library?

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+        gamma: Parameter determining maximum persistence value. Default is `1.0 / 128 * N_imgs / 5000`
+        i_max: Upper bound on the value of beta_1 to compute.
+    
+    Returns:
+      An array of size (i_max, ) containing RLT(i, 1, X, L)
+        for randomly sampled landmarks.
+    """
+    intervals, alpha_max = witness(X, sample_size=sample_size, gamma=gamma)
+    result = relative(intervals, alpha_max, i_max=i_max)
+    return result
```
snk4tr, line 151: The same as for the `rlts`: would transform it into a method

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+    intervals = st.persistence_intervals_in_dimension(1)
+    return intervals, alpha_max
+
+
+def rlt(
+        idx: int, X: np.ndarray, sample_size: int = 64, gamma: Optional[float] = None, i_max: int = 100) -> np.ndarray:
+    """Implements Algorithm 1 for one sample of landmarks.
+ 
+    Args:
+        idx : Used for multiprocessing.Pool to work correctly
```
snk4tr, line 139: Currently unused. May be this PR is not full?

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+    b = a[np.arange(np.shape(a)[0])[:, np.newaxis], idx]
+    distances = np.dstack([idx, b])
+    return distances, max_dist
+
+
+def witness(X, sample_size: int = 64, gamma: Optional[float] = None):
+    """Compute the persistence intervals for the dataset X using the witness complex.
+
+    Args:
+        X: Array of shape (N_samples, data_dim) representing the dataset.
```
snk4tr, line 96: I would propose to use lower case variable names when possible to satisfy Goodle style guide

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    a = cdist(W, L)
+    max_dist = np.max(a)
+    idx = np.argsort(a)
+    b = a[np.arange(np.shape(a)[0])[:, np.newaxis], idx]
+    distances = np.dstack([idx, b])
+    return distances, max_dist
+
+
+def witness(X, sample_size: int = 64, gamma: Optional[float] = None):
```
snk4tr, line 92: Please add returned data type in form of `Tuple[type1, type2]`

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+            # Count how many intervals contain midpoint
+            if midpoint >= interval[0] and midpoint < interval[1]:
+                s = s + 1
+        if (s < i_max):
+            rlt[s] += (switch_points[i + 1] - switch_points[i])
+
+    return rlt / alpha_max
+
+
+def lmrk_table(W: np.ndarray, L: np.ndarray):
```
snk4tr, line 71: Please add returned data type

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+
+    persistence_intervals = []
+    # If for some interval we have that it persisted up to np.inf
+    # we replace this point with alpha_max.
+    for interval in intervals:
+        if not np.isinf(interval[1]):
+            persistence_intervals.append(list(interval))
+        elif np.isinf(interval[1]):
+            persistence_intervals.append([interval[0], alpha_max])
```
snk4tr, line 42: Code style:
1. Try to not create conditions with reverse logic: replace `if not something` with `if something`. It makes understanding easier and improves readability
2. Try to avoid `elif` when `else` is applicable. It will explicitly show that you cover all possible variants of conditions
3. (Optional) Try to avoid `else` where it is not necessary. For example in this case you can replace it with `continue` within `if` statement. It reduces nesting, which simplifies code.

```suggestion
        if np.isinf(interval[1]):
            persistence_intervals.append([interval[0], alpha_max])
        else:
            persistence_intervals.append(list(interval))
            
```

or

```suggestion
        if np.isinf(interval[1]):
            persistence_intervals.append([interval[0], alpha_max])
            continue
        persistence_intervals.append(list(interval))
            
```

photosynthesis_metrics/gs.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # If there are no intervals in H1 then we always observed 0 holes.
+    if len(persistence_intervals) == 0:
+        rlt = np.zeros(i_max)
+        rlt[0] = 1.0
+        return rlt
```
snk4tr, line 48: `rlt` matches function's name, which can be confusing. Try to use a different name if possible.
Also `rlt` is a bad name in general because it is hard to understand what stands behind it. Try to change this name if possible

zakajd:
Okey, so I changed a couple things and now it should be good.
Here is some notes on improvements:
1. `np.random.seed()` in `witness` function is needed because otherwise numpy generates simular numbers in different threads. It's a common, but not well-known problem of using miltiprocessing.Pool()
2. @denproc  suggested to use torch.cdist, but it can be [**40x times**](https://github.com/pytorch/pytorch/issues/15253) slower, than scipy version. I tried to get rid of numpy and rewritten everything on pure PyTorch as suggested in #56. Result: **3.5** _increase_ in computation time, mainly because of cdist function (and even using [this code](https://github.com/pytorch/pytorch/issues/15253#issuecomment-491467128) didn't help much). So rolled everything back. 

Review it again and let's merge

tests/test_gs.py
zakajd: Everything is done in `BaseFeatreMetric` class. So there is no need to add more test here.

zakajd:
@denproc @snk4tr 
Hey, let's close this PR. Please review and merge it if everything is OK

PR #42, opened 2020-05-13
Feature/issue#23: Inception Score

### LINKED ISSUE #23: New measure: inception score
[Inception score](https://arxiv.org/abs/1606.03498) is a popular measure in the GAN world, hence need to be added.

Implement from scratch or adapt external implementation. Support `BaseFeatureMetric` interface.

zakajd:
Closes #23

Difference from existing open-source implementations:
1. IS computation is written on pure PyTorch without any external dependencies (like `numpy` or `scipy`). 
2. It's more robust, than naive Numpy implementations, which tend to produce `nan` values instead of `inf` and has approximately the same speed as version using `scipy`.


zakajd:
This implementation is similar to the one proposed in [A Note on the Inception Score ](https://arxiv.org/pdf/1801.01973.pdf). The difference from original paper is in reduction of scores for individual images. It calculates the exponential of the mean of the KL-divergence, not mean of the exponentials.

I added test on a subset of CIFAR10 (10.000 images), which gives results close to the paper. It's computed only if GPU is available and takes ~15 seconds. 

Review it and let's merge if everything OK

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,93 @@
...
+        super(IS, self).__init__()
+
+        self.num_splits = num_splits
+        self.ret_target = ret_target
+        self.ret_var = ret_var
+
+    def forward(
+        self, predicted_features: torch.Tensor, target_features: Optional[torch.Tensor] = None) -> torch.Tensor:
+        # Sanity check for input.
+        _validate_features(predicted_features, predicted_features if target_features is None else target_features)
```
denproc, line 54: See next comment at line 66.
```suggestion
        _validate_features(predicted_features, target_features if self.ret_target else predicted_features)
```

tests/test_is.py
```diff
@@ -0,0 +1,147 @@
+from typing import Optional
+
+import torch
+import pytest
+import torchvision
+import numpy as np
+from scipy.stats import entropy
```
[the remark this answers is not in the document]
zakajd: It's used only to compare that our result is close to other implementations. IS implementation is written on PyTorch.

We can add `scipy` as extra dependency only for development branch.

photosynthesis_metrics/__init__.py
```diff
@@ -3,3 +3,4 @@
 from .fid import FID
 from .kid import KID
 from .tv import TVLoss, total_variation
+from .inception_score import IS
```
snk4tr, line 6: Why the file with Inception Score is named `inception_score.py` but not `is.py`?

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.compute_metric(predicted_features, target_features)
```
snk4tr, line 55: As I remember, during one of the last refactorings, the `forward` method was unified between all feature-based metrics and moved to the `BaseFeatureMetric` class. Why did you decide to override it here?

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,93 @@
...
+        predicted_is = self.logits_to_score(predicted_features)
+        if self.ret_target:
+            target_is = self.logits_to_score(target_features)
+            return predicted_is, target_is
+        
+        return predicted_is
+
+    def logits_to_score(self, logits):
+        r"""Computes mean KL divergence.
+        """
```
snk4tr, line 74:
```suggestion
        r"""Computes mean KL divergence."""
```

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,93 @@
...
+        N = logits.size(0)
+        probas = F.softmax(logits)
+        split_scores = []
+        for i in range(self.num_splits):
+            part = probas[i * (N // self.num_splits): (i+1) * (N // self.num_splits), :]
+            p_y = torch.mean(part, dim=0)
+            scores = []
+            for k in range(part.shape[0]):
```
snk4tr, line 82: What is `part` here and what is its 0th dimension? Could you please introduce a variable with some descriptive name for easier understanding of what you are iterating over.

tests/test_is.py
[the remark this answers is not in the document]
snk4tr: Let's try to get rid of dependencies such as `scipy` as much as  possible

tests/test_is.py
```diff
@@ -0,0 +1,147 @@
...
+
+@pytest.fixture(scope='module')
+def features_prediction_beta() -> torch.Tensor:
+    m = torch.distributions.Beta(2, 2)
+    return m.sample((1000, 20))
+
+
+# @pytest.fixture(scope='module')
+# def features_prediction_constant() -> torch.Tensor:
+#     return torch.ones(1000, 20)
```
snk4tr, line 49: It would be good to remove commended code before merge into master

tests/test_is.py
```diff
@@ -0,0 +1,147 @@
...
+                break
+
+    metric = IS(ret_var=True)
+    target_features = torch.cat(target_features, dim=0)
+    mean, variance = metric(target_features, target_features)
+    # # Values from paper https://arxiv.org/pdf/1801.01973.pdf
+    # # CIFAR10 train: 9.737±0.148
+    mean_diff, var_diff = abs(mean - 9.737), abs(variance - 0.148)
+    assert (mean_diff <= 1.0) and (var_diff <= 0.5), \
+        f'Mean and var not close to paper. Mean diff: {mean_diff}, var diff: {var_diff}'
```
snk4tr, line 145: Nice!

tests/test_is.py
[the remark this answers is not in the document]
zakajd: > Let's try to get rid of dependencies such as `scipy` as much as possible
I can delete compatibility test with other implementations and leave only comparison with values from the paper. OK?

photosynthesis_metrics/inception_score.py
zakajd: Inception Score doesn't require 2 datasets to be computed. It can be measured only for 1 input and I tried to take this scenario into account

photosynthesis_metrics/__init__.py
zakajd: Because then in `__init__.py` when I try to write
`from .is import IS, inception_score` Python throws an error for incorrect syntax.

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,94 @@
...
+        Returns:
+            predicted_score: Scalar value of IS for predicted images features.
+            target_score: Scalar value of IS for target images features if `ret_target` is True.
+        """
+        predicted_is, _ = inception_score(predicted_features, num_splits=self.num_splits)
+        target_is, _ = inception_score(target_features, num_splits=self.num_splits)
+        if self.distance == 'l1':
+            return torch.dist(predicted_is, target_is, 1)
+        elif self.distance == 'l2':
+            return torch.dist(predicted_is, target_is, 2)
```
snk4tr, line 94: What happens if `self.distance` is neither 'l1' nor 'l2'? I think there should be `else` with `ValueError` and error message that explains that invalid distance type is provided

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,94 @@
...
+
+class IS(BaseFeatureMetric):
+    r"""Creates a criterion that measures difference of Inception Score between two datasets.
+    IS is computed separatly for predicted and target features and expects raw InceptionV3 model logits as inputs.
+
+    Args:
+        predicted_features (torch.Tensor): Low-dimension representation of predicted image set.
+            Required to have shape (N_pred, encoder_dim)
+        target_features (torch.Tensor): Low-dimension representation of target image set.
+            Required to have shape (N_targ, encoder_dim)
```
snk4tr, line 62: If it is required for `predicted features` and `target_features` to be 2D than it has to be validated. Failing fast with descriptive error message is always a good thing

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,94 @@
...
+
+from photosynthesis_metrics.base import BaseFeatureMetric
+
+
+def inception_score(features: torch.Tensor, num_splits: int = 10):
+    r"""Compute Inception Score for a list of image features.
+    Expects raw logits from Inception-V3 as input.
+
+    Args:
+        features (torch.Tensor): Low-dimension representation of image set. Shape (N, encoder_dim).
```
snk4tr, line 25: Assert correct number of dimensions for this parameter

photosynthesis_metrics/inception_score.py
```diff
@@ -0,0 +1,94 @@
...
+            num_splits: Number of parts to devide features.
+                IS is computed for them separatly and results are then averaged.
+            distance: How to measure distance between scores. One of {`l1`, `l2`}. Default: `l1`.
+        """
+        super(IS, self).__init__()
+        self.num_splits = num_splits
+        self.distance = distance
+
+    def compute_metric(
+            self, predicted_features: torch.Tensor, target_features: torch.Tensor) -> torch.Tensor:
```
snk4tr, line 82: Add docstring for `predicted_features` and `target_features`
Also  add assertion for the correctness of number of dimensions

tests/test_is.py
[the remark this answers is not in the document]
snk4tr: As we discussed verbally, it is better to leave test as long as it doesn't introduce extra dependencies for the library

photosynthesis_metrics/inception_score.py
zakajd: Validation checks are done in `forward` pass for all feature metrics.

photosynthesis_metrics/inception_score.py
zakajd: Docstring done. 
Assertions are made in `forward`, before calling `compute_metric` function.

snk4tr:
Waiting for approval from @denproc to merge

PR #45, opened 2020-05-13
bug/issue#44

zakajd:


zakajd:
Actual speed increase for tests is even larger. 
**12m 32 sec** -> **36 sec**

PR #47, opened 2020-05-13
feature/issue#33: Added Visual Information Fidelity metric

zakajd:
Implementation of an algorithm from paper https://ieeexplore.ieee.org/abstract/document/1576816

### DIFF
--- a/photosynthesis_metrics/vif.py
+++ b/photosynthesis_metrics/vif.py
@@ -6,7 +6,6 @@ References:
     https://ieeexplore.ieee.org/abstract/document/1576816/
 """
 import torch
-import numpy as np
 from torch.nn.modules.loss import _Loss
 import torch.nn.functional as F
 
@@ -51,6 +50,7 @@ def vif_p(prediction: torch.Tensor, target: torch.Tensor,
         See https://live.ece.utexas.edu/research/Quality/VIF.htm for details.
         
     """
+    assert prediction.dim() == 4, f'Expected 4D tensor, got {prediction.size()}'
     if data_range == 255:
         prediction = prediction / 255.
         target = target / 255.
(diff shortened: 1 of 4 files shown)


zakajd:
Here are some notes about this implementation of VIF:
1. Originally VIF is computed for wavelet bands, but later authors released code on MATLAB to compute VIF in pixel domain. I adopted it in PyTorch, thats why main function is called `vif_p`.
2. VIF works only on greyscale images, so I added conversion RGB -> YCbCr and take luminance component if input images have 3 channels.
3. VIF is metric, that should be maximised (like SSIM), so in order to be used as loss I return `1 - VIF` value.

photosynthesis_metrics/vif.py
```diff
@@ -0,0 +1,138 @@
...
+        prediction, target = _adjust_dimensions(x=prediction, y=target)
+        _validate_input(x=prediction, y=target)
+
+        return self.compute_metric(prediction, target)
+
+    def compute_metric(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+
+        score = vif_p(prediction, target, sigma_n_sq=self.sigma_n_sq, data_range=self.data_range)
+        # Make sure value to be in [0, 1] range and convert to loss
+        loss = 1 - torch.clamp(torch.mean(score), 0, 1)
```
denproc, line 137: In case of `SSIM` and `MS-SSIM`, we return the value, which should be maximised. What best practice should we use? Make all losses to be minimised or leave it for user?

tests/test_vif.py
```diff
@@ -0,0 +1,63 @@
...
+        VIFLoss()
+    except Exception as e:
+        pytest.fail(f"Unexpected error occurred: {e}")
+
+
+def test_vif_loss_zero_for_equal_tensors(prediction: torch.Tensor):
+    loss = VIFLoss()
+    target = prediction.clone()
+    measure = loss(prediction, target)
+    assert measure <= 1e-6, f'VIF for equal tensors must be 0, got {measure}'
```
denproc, line 63: Add tests to check input type, size and number of dimensions.

photosynthesis_metrics/vif.py
```diff
@@ -0,0 +1,138 @@
...
+    Returns:
+        VIF: Index of similarity betwen two images. Usually in [0, 1] interval.
+            Can be bigger than 1 for predicted images with higher contrast than original one.
+    Note:
+        In original paper this method was used for bands in discrete wavelet decomposition.
+        Later on authors released code to compute VIF approximation in pixel domain.
+        See https://live.ece.utexas.edu/research/Quality/VIF.htm for details.
+        
+    """
+    if data_range == 255:
```
denproc, line 54: Current `_validate_input` allows 5D tensor of `(N,C,H,W,2)` size. 
```suggestion
    assert prediction.dim() == 4, f'Expected 4D tensor, got {prediction.size()}'
    if data_range == 255:
```

photosynthesis_metrics/vif.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+import torch
+import numpy as np
+from torch.nn.modules.loss import _Loss
+import torch.nn.functional as F
+
+from photosynthesis_metrics.utils import _adjust_dimensions, _validate_input
+
+
+def _gaussian_kernel2d(kernel_size: int = 5, sigma: float = 2.0) -> torch.Tensor:
```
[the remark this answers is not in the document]
zakajd: We can open an issue, but for now I would suggest to leave it like that. In SSIM 1-D kernels are used, and here they are 2-D. So there are not exactly the same.

photosynthesis_metrics/vif.py
zakajd: Yeah. I thin loss is intended to be minimised. We can either return `1 - metric` or write in the docs that user need to do that themselves.

photosynthesis_metrics/vif.py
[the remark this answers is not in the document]
denproc: The ideas are actually the same, in case of SSIM we apply 1d kernel for X and Y, while here we apply 2d kernel. 
Nevertheless, I agree that we could consider it as additional issue/feature.

PR #55, opened 2020-05-20
Enhancement of SSIM, MS-SSIM and TV

### LINKED ISSUE #6: Add support of 5D tensors in SSIM and MS-SSIM
Current implementation supports only 2, 3, and 4 dimensional tensors. However, in some domains like medical imaging it may be useful to be able to work with 5D tensors out of the box.

Fix the `_compute_ssim` and `_compute_multi_scale_ssim` functions so they can handle 5D.

Fix top-level functions to make then iterate over the extra dimension and then reduce.

**Additional context**
-

### LINKED ISSUE #8: Unify SSIM/MS-SSIM and their losses in terms of supported tensor sizes
Currently SSIM/MS-SSIM losses try to work with 2, 3, 4-dimensional tensors, but corresponding measures support only 4D. Unify that.

Provide interface for the unification.

None

**Additional context**
None

### LINKED ISSUE #51: Solve adjust dimentions/validate input problem
Currently, not feature-based methods perform two steps before actual computation of metrics: adjustment of dimensions (which is meant to make processing of tensors with different number of dimensions possible) and validation of input (which is meant to validate correctness of user's input). However, their existing implementations contradict to each other: the validation function validates that user's input is 4-dimensional while `adjust_dimention` changes user's input to be 4-dimensional. This behaviour is redundant and introduces extra confusion.

* Remove check that user's input is 4-dimensional
* Make sure that `adjust_dimentions` functionality works as expected (introduce extra tests)
* Implement all above for all not feature-based metrics

**Additional context**
It is important to keep in mind #6. We would like to work not only with 4-dimensional tensors because in some applications like medical imaging 5D tensors are used. It is also sometimes convenient for a user to pass 2D or 3D tensors and not perform adjustment themselves.

denproc:
Closes #6 
Closes #8 
Closes #51 

### Major
- SSIM, MS-SSIM and TV features were upgraded to support 2D, 3D or 4D input images;
- Tests to check 2D, 3D and 4D support for SSIM, MS-SSIM and TV;
- SSIM and MS-SSIM were upgraded to support 5D input images of `(N,C,H,W,2)` size, where the last dimension is responsible for complex values;
- Tests to validate 5D support.
### Minor
- MS-SSIM was upgraded to use custom scale weights of different size;
- Minor changes to improve readability and to optimise code.

### DIFF
--- a/photosynthesis_metrics/ssim.py
+++ b/photosynthesis_metrics/ssim.py
@@ -50,7 +50,6 @@ def ssim(x: torch.Tensor, y: torch.Tensor, kernel_size: int = 11, kernel_sigma: 
                                  y=y,
                                  kernel=kernel,
                                  data_range=data_range,
-                                 size_average=False,
                                  full=True,
                                  k1=k1,
                                  k2=k2)
@@ -190,7 +189,6 @@ class SSIMLoss(_Loss):
             y=target,
             kernel=kernel,
             data_range=self.data_range,
-            size_average=False,
             full=False,
             k1=self.k1,
             k2=self.k2
@@ -442,8 +440,6 @@ def _ssim_per_channel(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, da
             y: Batch of images, (N,C,H,W).
             kernel: 1-D gauss kernel.
             data_range: Value range of input images (usually 1.0 or 255).
-            size_average: If size_average=True, ssim of all images will be averaged as a scalar.
-            full: Return sc or not.
             k1: Algorithm parameter, K1 (small constant, see [1]).
             k2: Algorithm parameter, K2 (small constant, see [1]).
                 Try a larger K2 constant (e.g. 0.4) if you get a negative or NaN results.
@@ -494,7 +490,6 @@ def _compute_ssim(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, data_r
         y: Batch of images, (N,C,H,W).
         kernel: 1-D gauss kernel.
         data_range: Value range of input images (usually 1.0 or 255).
-        size_average: If size_average=True, ssim of all images will be averaged as a scalar.
         full: Return sc or not.
         k1: Algorithm parameter, K1 (small constant, see [1]).
         k2: Algorithm parameter, K2 (small constant, see [1]).
@@ -506,12 +501,8 @@ def _compute_ssim(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, data_r
 
     ssim_map, cs_map = _ssim_per_channel(x=x, y=y, kernel=kernel, data_range=data_range, k1=k1, k2=k2)
 
-    if size_average:
-        ssim_val = ssim_map.mean()
-        cs = cs_map.mean()
-    else:
-        ssim_val = ssim_map.mean(-1)
-        cs = cs_map.mean(-1)
+    ssim_val = ssim_map.mean(-1)
+    cs = cs_map.mean(-1)
 
     if full:
         return ssim_val, cs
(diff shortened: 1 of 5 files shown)


photosynthesis_metrics/ssim.py
```diff
@@ -142,7 +140,7 @@ class SSIMLoss(_Loss):
     def __init__(self, kernel_size: int = 11, kernel_sigma: float = 1.5, k1: float = 0.01, k2: float = 0.03,
                  size_average: Optional[bool] = None, reduce: Optional[bool] = None,
```
zakajd, line 141: Why do we need arguments `size_average`, `reduce` and `reduction` all at once? It matches old-style PyTorch functionality, but now only `reduction` is used and 2 other options are deprecated.
I propose to get rid of them too and left only `reduction` parameter which takes a string as input.

photosynthesis_metrics/ssim.py
```diff
@@ -362,7 +369,7 @@ def __init__(self, kernel_size: int = 11, kernel_sigma: float = 1.5, k1: float =
         if scale_weights is None:
             scale_weights_from_ms_ssim_paper = [0.0448, 0.2856, 0.3001, 0.2363, 0.1333]
```
zakajd, line 370: All this `self.scale_weights_tensor` and `scale_weights_from_ms_ssim_paper` variables look redundant.
Maybe we can change this to be like:
```python
if scale_weights is None:
    # Scale weight from MS-SSIM paper
    scale_weights = [0.0448, 0.2856, 0.3001, 0.2363, 0.1333]
self.scale_weights = torch.tensor(scale_weights)
#no self.scale_weights_tensor attribute
```
And then use `self.scale_weights` everywhere?

photosynthesis_metrics/utils.py
```diff
@@ -27,18 +27,21 @@ def _validate_input(
...
+    assert x.size() == y.size(), f'Input images must have the same dimensions, got {x.size()} and {y.size()}.'
+    if x.dim() == 5:
+        assert x.size(-1) == 2, f'Expected Complex 5D tensor with (N,C,H,W,2) size, got {x.size()}'
     if kernel_size is not None:
         assert kernel_size % 2 == 1, f'Kernel size must be odd, got {kernel_size}.'
     if scale_weights is not None:
         assert isinstance(scale_weights, (list, tuple, torch.Tensor)), \
             f'Scale weights must be of type list, tuple or torch.Tensor, got {type(scale_weights)}.'
-        assert len(scale_weights) == 5, f'Scale weights collection must contain 5 values, got {len(scale_weights)}.'
+        assert (torch.Tensor(scale_weights).dim() == 1), \
```
zakajd, line 39: It's better to use `torch.tensor` function instead of `torch.Tensor` for converting into tensors, because output dtype is same as inputs. In this particular case result will be the same `torch.float32`, but sometimes it can lead to types mismatch.
See https://discuss.pytorch.org/t/difference-between-torch-tensor-and-torch-tensor/30786 for details

photosynthesis_metrics/ssim.py
denproc: I suppose `scale_weights_from_ms_ssim_paper` is just to improve readability.
Regarding `self.scale_weights` and `self. self.scale_weights_tensor`, I'll reduce it to just `self.scale_weights` usage.

photosynthesis_metrics/ssim.py
denproc: It seems to be redundant, but it allows us to support legacy. ¯\_(ツ)_/¯
In addition, we have similar parameters @ `tv.py`, where they are used differently. I would suggest to analyse it as a separate issue (Unify interfaces of `SSIMLoss`, `MultiScaleLoss` and 'TVLoss').

snk4tr:
Gonna have a look tomorrow

photosynthesis_metrics/ssim.py
snk4tr: Agree, it is out of the scope of this PR

denproc:
Waiting for changes to be introduces in #58

denproc:
Updated the PR with new changes from master. Ready for review.

zakajd:
Will review tomorrow.

denproc:
@snk4tr @zakajd , let's do it in the following way:
Current implementation (including all children of `_Loss`) doesn't support differently sized input tensors. `_validate_input` checks tensors to be of the same size. I'll add the tests for SSIM to catch the required alerts, but user will not be able to pass tensors with different number of dimensions. 
Agree?

denproc:
Ready for review, @snk4tr @zakajd. Let's finally nail this PR :)

PR #58, opened 2020-05-23
Bug/issue#57

smikhai1:
I changed the TVLoss computation to be based on only predicted image, as it's [used](https://arxiv.org/abs/1603.08155) as a regularizer applied to the predicted images during NN training.

As the functions `_adjust_dimensions` and `_validate_inputs` from `utils` module only support two input tensors, I added their counterparts `_adjust_tensor_dimensions` and `_validate_input` directly in `tv` module. In the future, they may be combined with the corresponding functions from utils.

The README file and tests are changed to be up to the fix.

### DIFF
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -23,7 +23,7 @@ def _adjust_tensor_dimensions(x: torch.Tensor):
 def _validate_input(x: torch.Tensor) -> None:
     """Validates input tensor"""
     assert isinstance(x, torch.Tensor), f'Input must be a torch.Tensor, got {type(x)}.'
-    assert len(x.shape) == 4, f'Input image must be 4D tensor, got image of shape {x.shape}.'
+    assert x.dim() == 4, f'Input image must be 4D tensor, got image of shape {x.shape}.'
 
 
 def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type: str = 'l2') -> torch.Tensor:
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -5,7 +5,25 @@ r"""Implemetation of Total Variation metric, based on article
 import torch
 from torch.nn.modules.loss import _Loss
 
-from photosynthesis_metrics.utils import _adjust_dimensions, _validate_input
+
+def _adjust_tensor_dimensions(x: torch.Tensor):
+    r"""Expands input tensor dimensions to 4D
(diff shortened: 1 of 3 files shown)


snk4tr:
@zakajd @denproc can you have a look and comment?

tests/test_tv.py
```diff
@@ -19,31 +19,24 @@ def test_tv_works(prediction: torch.Tensor) -> None:
...
-
 def test_tv_for_known_answer():
     # Tensor with `l1` TV = (10 - 1) * 2  * 2 = 36
     prediction = torch.eye(10).reshape((1, 1, 10, 10))
-    # Tensor with TV = 0
-    target = torch.zeros((1, 1, 10, 10))
     loss = TVLoss(reduction_type='l1')
-    measure = loss(prediction, target)
+    measure = loss(prediction)
     assert measure == 36., f'TV for this tensors must be 36., got {measure}'
```
snk4tr, line 42: I think this test will break when changes are incorporated. Since the purpose of this test is to verify correctness of computations, please introduce another value that you are confident in or compare the result with some other trustworthy implementation.

smikhai1: I checked the tests from `test_tv.py` on my machine and all of them were passed successfully

smikhai1: @snk4tr btw, how to turn on the Actions pipeline?

snk4tr: @smikhai1 first, check that Actions are not disabled in your fork of the repo. If they are, try to enable them. Note that testing and linting pipelines are triggered with new pushed commits 
image

photosynthesis_metrics/tv.py
```diff
@@ -98,33 +109,26 @@ def __init__(self, size_average: bool = True, reduction_type: str = 'l2', reduct
...
             prediction: Tensor of prediction of the network.
-            target: Reference tensor.
 
         Returns:
             Value of TV loss to be minimized.
         """
-        prediction, target = _adjust_dimensions(x=prediction, y=target)
-        _validate_input(x=prediction, y=target)
+        prediction = _adjust_tensor_dimensions(prediction)
+        _validate_input(prediction)
```
denproc, line 122: Let's change the order in order to validate input first and then adjust the dimensions. Current implementation makes `isinstance(x, torch.Tensor)` at  `_validate_input` pointless.

smikhai1: yeah, it's a good point!

PR #62, opened 2020-05-24
Make workflows run from PRs

### LINKED ISSUE #61: Workflows are not running in PRs
**Describe the bug**
If PR is introduced from a fork, the results of workflows are not visible in the PR.

**To Reproduce**
1. Fork the repo
2. Introduce your changes
3. Make PR
4. See that no workflows were run even though they were

**Expected behavior**
Workflows are running the same way as they do in case of PRs from branches in the repo

snk4tr:
Closes #61 

## Proposed Changes

  - Changes files for GitHub Actions (workflows) to make actions trigger not only on pushes but also on PRs

### DIFF
--- a/.github/workflows/ci-linting.yml
+++ b/.github/workflows/ci-linting.yml
@@ -1,6 +1,6 @@
 name: flake-8 style check
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


denproc:
Do not worry, next events in this pull request are to check the PR.

denproc:
Well, the workflows are triggered twice because two independent events happened. What are the thought about leaving the CI only with trigger on `pull_request` event? 
Testing could be done locally before pushing, but it is essential to show that all checks were passed at PR.

snk4tr:
Agree. One also can create draft PR and checks can still run remotely if needed. Removing hook of push

PR #63, opened 2020-05-25
Bug/trigger ci

denproc:
**Changes**
- Trigger CI not only on PR but also on push in `master` branch.


snk4tr:
Nice one! This what we actually need :)

Issue #64, opened 2020-05-25
Clean up the redundant dependencies for loss initialisation.


denproc:
**Problem.**
SSIM and MS-SSIM use set of parameters (`size_average`, `reduction`, `reduce`), which introduce same functionality.
`TVLoss` uses `size_average`, but this parameter is not needed.

**Solution.**
Remove redundant parameters and perform required optimisation of code.

Raised at #55  by @zakajd.

PR #66, opened 2020-05-26
bug(tv): reduction was handled to return proper values

### LINKED ISSUE #65: Same values for `mean` and `sum` reductions of TV loss.
**Describe the bug**
Reductions at `TVLoss` return the same value, when default parameters are used to initialise `TVLoss()`

**To Reproduce**
```python
from photosynthesis_metrics import TVLoss
import torch
torch.manual_seed(42)
a = torch.rand(3,3,256,256)
print(f'mean reduction: {TVLoss(reduction='mean')(a)}')
print(f'sum reduction: {TVLoss(reduction='sum')(a)}')
```
Output
```python
>>> mean reduction: 255.3401
>>> sum reduction: 255.3401
```

**Expected behavior**
Output
```python
>>> mean reduction: 255.3401
>>> sum reduction: 766.0204
```

**Solution**
- Get rid of `size_average` parameter at `TVLoss`;
- Call `total_variation` with `size_average=False` at `TVLoss().forward()`

denproc:
Closes #65 issue. Related to #64.

**Proposed changes:**
- `size_average` parameter was eliminated from `TVLoss` initialisation;
- `total_variation` is called with `size_average=False`;
- `reduction` is checked to be `"none"|"mean"|"sum"`.

### DIFF
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -57,9 +57,9 @@ def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type: 
         raise ValueError("Incorrect reduction type, should be one of {'l1', 'l2', 'l2_squared'}")
 
     if size_average:
-        return tv_val.mean()
-    else:
-        return tv_val
+        return tv_val.mean(dim=0)
+    
+    return tv_val
 
 
 class TVLoss(_Loss):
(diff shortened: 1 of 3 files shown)


denproc:
If we would like to postpone the bug fix, we can do it here, but after merging #55.

zakajd:
> If we would like to postpone the bug fix, we can do it here, but after merging #55.

Yes, let's merge #55  first.

denproc:
Ready for review. Now it also resolves #64

photosynthesis_metrics/tv.py
```diff
@@ -23,7 +23,7 @@ def _adjust_tensor_dimensions(x: torch.Tensor):
 def _validate_input(x: torch.Tensor) -> None:
     """Validates input tensor"""
     assert isinstance(x, torch.Tensor), f'Input must be a torch.Tensor, got {type(x)}.'
-    assert 1 < x.dim() < 5, f'Input image must be 4D tensor, got image of shape {x.size()}.'
```
snk4tr, line 26: Good catch!

Issue #35, opened 2020-05-07
Asserting minimal image resolution for MS-SSIM computation


snk4tr:
**Is your feature request related to a problem? Please describe.**
Currently is low resolution image/set of images is passed to the MultiScaleSSIMLoss it crashes with the default PyTorch stack trace, which can be hard to interpret

**Describe the solution you'd like**
Find out the minimal allowed resolution and introduce an assertion/check of image resolution before it is passed to the MS-SSIM computation code.

PR #68, opened 2020-05-26
Feature: GMSD and MultiScale GMSD

### LINKED ISSUE #67: Metric: Gradient Magnitude Similarity and it's multi-scale version
New metrics to measure full-reference image quality based on gradients similarity. 
It was shown to improve Super-Resolution quality compared to usage of only MSE loss. 

Add Gradient magnitude similarity deviation (GMSD) https://arxiv.org/pdf/1308.3052.pdf
And Multiscale GMSD https://ieeexplore.ieee.org/document/7952357

zakajd:
Closes #67

### DIFF
--- a/photosynthesis_metrics/gmsd.py
+++ b/photosynthesis_metrics/gmsd.py
@@ -232,7 +232,7 @@ class MultiScaleGMSDLoss(_Loss):
         score = ms_gmds_val
         
         if self.chromatic:
-            assert prediction.size(1) == 3, f"Chromatic component can be computed only for RGB images!"
+            assert prediction.size(1) == 3, "Chromatic component can be computed only for RGB images!"
             
             # Convert to YIQ color space https://en.wikipedia.org/wiki/YIQ
             iq_weights = torch.tensor([[0.5959, -0.2746, -0.3213], [0.2115, -0.5227, 0.3112]]).t()
--- a/photosynthesis_metrics/gmsd.py
+++ b/photosynthesis_metrics/gmsd.py
@@ -8,6 +8,7 @@ Reference:
     http://www.cse.ust.hk/~psander/docs/gradsim.pdf
 
 """
+import torch
 from typing import Optional, Union, Tuple, List
 
 import torch.nn.functional as F
@@ -23,8 +24,8 @@ def _prewitt_filter() -> torch.Tensor:
     return kernel
 
 
-def _compute_gmsd(prediction: torch.Tensor, target: torch.Tensor, 
-         reduction: Optional[str] = 'mean') -> torch.Tensor:
+def _gmsd(prediction: torch.Tensor, target: torch.Tensor,
+          reduction: Optional[str] = 'mean') -> torch.Tensor:
     r"""Compute Gradient Magnitude Similarity Deviation
     Both inputs supposed to be in range [0, 1] with RGB order.
     Args:
@@ -42,7 +43,7 @@ def _compute_gmsd(prediction: torch.Tensor, target: torch.Tensor,
         https://arxiv.org/pdf/1308.3052.pdf
     """
     # Constant for numerical stability
-    EPS : float = 1e-4
+    EPS: float = 1e-4
 
     # Convert RGB image to YCbCr and take luminance: Y = 0.299 R + 0.587 G + 0.114 B
     num_channels = prediction.size(1)
@@ -81,9 +82,10 @@ def _compute_gmsd(prediction: torch.Tensor, target: torch.Tensor,
         return gmsd.sum()
     else:
         return gmsd
-    
+
+
 class GMSDLoss(_Loss):
-    r"""Creates a criterion that measures Gradient Magnitude Similarity Deviation 
+    r"""Creates a criterion that measures Gradient Magnitude Similarity Deviation
     between each element in the input and target.
 
     Args:
@@ -136,7 +138,7 @@ class GMSDLoss(_Loss):
         prediction = F.avg_pool2d(prediction, kernel_size=2, stride=2, padding=padding)
         target = F.avg_pool2d(target, kernel_size=2, stride=2, padding=padding)
         
-        score = _compute_gmsd(
+        score = _gmsd(
             prediction, target, reduction=self.reduction)
 
         return score
@@ -156,7 +158,7 @@ class MultiScaleGMSDLoss(_Loss):
             i.e., if for image x it holds min(x) = 0 and max(x) = 1, then data_range = 1.
             The pixel value interval of both input and output should remain the same.
         scale_weights: Weights for different scales. Must contain 4 floating point values.
-        chromatic: Flag to use MS-GMSDc algorithm from paper. 
+        chromatic: Flag to use MS-GMSDc algorithm from paper.
             It also evaluates chromatic components of the image. Default: True
         beta1: Algorithm parameter. Weight of chromatic component in the loss.
         beta2: Algorithm parameter. Small constant, see [1].
@@ -167,9 +169,9 @@ class MultiScaleGMSDLoss(_Loss):
             http://www.cse.ust.hk/~psander/docs/gradsim.pdf
     """
 
-    def __init__(self, reduction: str = 'mean', data_range: Union[int, float] = 1., 
+    def __init__(self, reduction: str = 'mean', data_range: Union[int, float] = 1.,
                  scale_weights: Optional[Union[Tuple[float], List[float]]] = None,
-                 chromatic: bool = True, beta1: float = 0.01, beta2: float = 0.32,
+                 chromatic: bool = False, beta1: float = 0.01, beta2: float = 0.32,
                  beta3: float = 15.) -> None:
         super().__init__()
 
@@ -219,7 +221,7 @@ class MultiScaleGMSDLoss(_Loss):
                 prediction = F.avg_pool2d(prediction, kernel_size=2, padding=padding)
                 target = F.avg_pool2d(target, kernel_size=2, padding=padding)
         
-            score = _compute_gmsd(prediction, target, reduction=None)
+            score = _gmsd(prediction, target, reduction=None)
             ms_gmds.append(score)
         
         # Concat results in different scales and multiply by weight
@@ -233,9 +235,9 @@ class MultiScaleGMSDLoss(_Loss):
             assert prediction.size(1) == 3, f"Chromatic component can be computed only for RGB images!"
             
             # Convert to YIQ color space https://en.wikipedia.org/wiki/YIQ
-            weights = torch.tensor([[0.5959, -0.2746, -0.3213],[0.2115, -0.5227, 0.3112]]).t()
-            prediction_iq = torch.matmul(prediction.permute(0, 2, 3, 1), weights).permute(0, 3, 1, 2)
-            target_iq = torch.matmul(target.permute(0, 2, 3, 1), weights).permute(0, 3, 1, 2)
+            iq_weights = torch.tensor([[0.5959, -0.2746, -0.3213], [0.2115, -0.5227, 0.3112]]).t()
+            prediction_iq = torch.matmul(prediction.permute(0, 2, 3, 1), iq_weights).permute(0, 3, 1, 2)
+            target_iq = torch.matmul(target.permute(0, 2, 3, 1), iq_weights).permute(0, 3, 1, 2)
             
             rmse_iq = torch.sqrt(torch.mean((prediction_iq - target_iq) ** 2, dim=[-1, -2]))
             rmse_chrome = torch.sqrt(torch.sum(rmse_iq ** 2, dim=1))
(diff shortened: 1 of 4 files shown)


photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+
+    # Compute GMSD
+    gmsd = torch.pow(gms - mean_gms, 2).mean(dim=[1, 2, 3], keepdims=True).sqrt()
+    
+    if reduction == 'mean':
+        return gmsd.mean()
+    elif reduction == 'sum':
+        return gmsd.sum()
+    else:
+        return gmsd
```
denproc, line 84: - Let's specify the dimension for the `mean` and `sum` operations. Therefore, we will have passive check that everything is calculated right and improve readability that we perform operation for batch dimension.
- Let's add check for specified `reduction` modes.
```suggestion
    if reduction == 'mean':
        return gmsd.mean(dim=0)
    elif reduction == 'sum':
        return gmsd.sum(dim=0)
    elif reduction != 'none':
        raise ValueError(f'Expected reduction modes are "mean"|"sum"|"none", got {reduction}')
    return gmsd
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+        r"""Computation of Gradient Magnitude Similarity Deviation (GMSD) as a loss function.
+
+        Args:
+            prediction: Tensor of prediction of the network.
+            target: Reference tensor.
+
+        Returns:
+            Value of GMSD loss to be minimized. 0 <= GMSD loss <= 1.
+        """
+        _validate_input(x=prediction, y=target, scale_weights=None)
```
denproc, line 205:
```suggestion
        _validate_input(x=prediction, y=target, scale_weights=self.scale_weights)
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+        scale_weights = self.scale_weights.to(prediction)
+        ms_gmds = []
+        for scale in range(4):
+            if scale > 0:
+                # Average by 2x2 filter and downsample
+                padding = (prediction.shape[2] % 2, prediction.shape[3] % 2)
+                prediction = F.avg_pool2d(prediction, kernel_size=2, padding=padding)
+                target = F.avg_pool2d(target, kernel_size=2, padding=padding)
+        
+            score = _gmsd(prediction, target, reduction=None)
```
denproc, line 224:
```suggestion
            score = _gmsd(prediction, target, reduction='none')
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+            ms_gmds.append(score)
+        
+        # Concat results in different scales and multiply by weight
+        ms_gmds_val = scale_weights.view(1, 4, 1, 1) * (torch.cat(ms_gmds, dim=1) ** 2)
+        # Sum and take sqrt per-image
+        ms_gmds_val = torch.sqrt(torch.sum(ms_gmds_val, dim=-1))
```
denproc, line 230: The sum is performed for `dim=4`, which equals 1. Shouldn't the operation be performed for `dim=(1,2,3)`?
```suggestion
        ms_gmds_val = torch.sqrt(torch.sum(ms_gmds_val, dim=(1, 2, 3)))
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+            gamma = 2 / (1 + self.beta2 * torch.exp(-self.beta3 * ms_gmds_val)) - 1
+            
+            score = gamma * ms_gmds_val + (1 - gamma) * self.beta1 * rmse_chrome
+            
+        if self.reduction == 'mean':
+            score = torch.mean(score)
+        elif self.reduction == 'sum':
+            score = torch.sum(score)
+
+        return score
```
denproc, line 253:
```suggestion
        if self.reduction == 'mean':
            score = torch.mean(score, dim=0)
        elif self.reduction == 'sum':
            score = torch.sum(score, dim=0)
        elif self.reduction != 'none':
            raise ValueError(f'Expected reduction modes are "mean"|"sum"|"none", got {self.reduction}')

        return score
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+    def compute_metric(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        
+        if self.data_range == 255:
+            prediction = prediction / 255.
+            target = target / 255.
+
+        # Average by 2x2 filter and downsample
+        padding = (prediction.shape[2] % 2, prediction.shape[3] % 2)
+        prediction = F.avg_pool2d(prediction, kernel_size=2, stride=2, padding=padding)
+        target = F.avg_pool2d(target, kernel_size=2, stride=2, padding=padding)
```
denproc, line 139: According to [last paper](http://www.cse.ust.hk/~psander/docs/gradsim.pdf) there is no need in downsample for GMSD. 
Also, it is stated that first GMDS at multiscale approach is calculated for original sized images. In current implementation, we do downsampling first, then calculate GMDS.

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+            
+        scale_weights = self.scale_weights.to(prediction)
+        ms_gmds = []
+        for scale in range(4):
+            if scale > 0:
+                # Average by 2x2 filter and downsample
+                padding = (prediction.shape[2] % 2, prediction.shape[3] % 2)
+                prediction = F.avg_pool2d(prediction, kernel_size=2, padding=padding)
+                target = F.avg_pool2d(target, kernel_size=2, padding=padding)
+
```
denproc, line 223: If there is no need in avgpooling at first stage, let's put `_gmsd` call first then pooling as it is done at MS-SSIM.

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
...
+        
+        score = ms_gmds_val
+        
+        if self.chromatic:
+            assert prediction.size(1) == 3, "Chromatic component can be computed only for RGB images!"
+            
+            # Convert to YIQ color space https://en.wikipedia.org/wiki/YIQ
+            iq_weights = torch.tensor([[0.5959, -0.2746, -0.3213], [0.2115, -0.5227, 0.3112]]).t()
+            prediction_iq = torch.matmul(prediction.permute(0, 2, 3, 1), iq_weights).permute(0, 3, 1, 2)
+            target_iq = torch.matmul(target.permute(0, 2, 3, 1), iq_weights).permute(0, 3, 1, 2)
```
denproc, line 240: The prediction and target are not equal to input ones as avgpooling was performed several times.

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,253 @@
+"""
```
denproc, line 1: Only classes were defined here. Should we define the methods for GMSD and MS-GMSD as we did for SSIM and MS-SSIM?

photosynthesis_metrics/gmsd.py
zakajd: See original paper for [GMSD]( https://arxiv.org/pdf/1308.3052.pdf) Section 3.B 
>In addition, as in the implementations of SSIM [8] and FSIM [7], the images r and d are first filtered by a 2×2 average filter, and then down-sampled by a factor of 2

They also do it in their MATLAB code. Another reason for this is weights for MS-GMSD, where the biggest weight in on scale 2, which is exactly blurred and downsampled image.

photosynthesis_metrics/gmsd.py
zakajd: See MS-GMSD paper, chapter 3.3
> It should be noted that chrominance dissimilarity is only evaluated at Scale M because the spatial resolution to color stimulation is much lower than luminance for vision perception.

photosynthesis_metrics/gmsd.py
zakajd: Why? The reason to have separate class for `SSIM` is that when it's used as a measure it's value _maximised_ and when used as a loss it's value _minimised_. Here for both cases value of GMSD is minimised, so class `GMSDLoss` can be used as a measure without any additional changes.

denproc: Well, actually no. As I understood, the solution was to return true value and describe it in the documentation that user should take it into account.

photosynthesis_metrics/gmsd.py
zakajd: We have different number of weights in MS-SSIM and MS-GMSD (5 and 4 respectively). So this change leads to assertion errors in `_validate_input` function.

denproc: `_validate_input` will support flexible number of weights after #55. It's already implemented.

denproc: May user use custom weights of different size? like `[0.4, 0.5, 0.3, 0.2, 0.1]`?

tests/test_gmsd.py
```diff
@@ -0,0 +1,111 @@
...
+    except Exception as e:
+        pytest.fail(f"Unexpected error occurred: {e}")
+
+
+def test_multi_scale_gmsd_small_for_tensors_with_contrast_difference():
+    loss = MultiScaleGMSDLoss(chromatic=True)
+    target = torch.ones(3, 3, 256, 256)
+    prediction = torch.zeros(3, 3, 256, 256)
+    measure = loss(prediction, target)
+    assert measure <= 1e-6, \
```
denproc, line 110:
```suggestion
    assert measure.abs() <= 1e-6, \
```

tests/test_gmsd.py
```diff
@@ -0,0 +1,111 @@
...
+        MultiScaleGMSDLoss()
+    except Exception as e:
+        pytest.fail(f"Unexpected error occurred: {e}")
+
+
+def test_multi_scale_gmsd_zero_for_equal_tensors(prediction: torch.Tensor):
+    loss = MultiScaleGMSDLoss()
+    target = prediction.clone()
+    measure = loss(prediction, target)
+    assert measure <= 1e-6, f'MultiScaleGMSD for equal tensors must be 0, got {measure}'
```
denproc, line 72:
```suggestion
    assert measure.abs() <= 1e-6, f'MultiScaleGMSD for equal tensors must be 0, got {measure}'
```

tests/test_gmsd.py
```diff
@@ -0,0 +1,111 @@
...
+        GMSDLoss()
+    except Exception as e:
+        pytest.fail(f"Unexpected error occurred: {e}")
+
+
+def test_gmsd_zero_for_equal_tensors(prediction: torch.Tensor):
+    loss = GMSDLoss()
+    target = prediction.clone()
+    measure = loss(prediction, target)
+    assert measure <= 1e-6, f'GMSD for equal tensors must be 0, got {measure}'
```
denproc, line 29:
```suggestion
    assert measure.abs() <= 1e-6, f'GMSD for equal tensors must be 0, got {measure}'
```

photosynthesis_metrics/gmsd.py
```diff
@@ -0,0 +1,256 @@
...
+        _validate_input(x=prediction, y=target, scale_weights=self.scale_weights)
+        prediction, target = _adjust_dimensions(x=prediction, y=target)
+
+        return self.compute_metric(prediction, target)
+    
+    def compute_metric(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        if self.data_range == 255:
+            prediction = prediction / 255.
+            target = target / 255.
+
```
denproc, line 215: Following #35, there should be an Assertion on the image sizes.

photosynthesis_metrics/gmsd.py
zakajd: Weights should sum to 1. A added weights normalization in `__init__`, so even when user writes `[2., 4., 1., 1.]` everything should work fine.

photosynthesis_metrics/gmsd.py
zakajd: Difference here is that for MS-SSIM minimal size is 32x32 and lot's of small datasets do not satisfy this condition. While for MS-GMSD minimal image size is 16x16. I doubt anyone will ever try to compute this metric between such small patches.

photosynthesis_metrics/gmsd.py
denproc: The point is to use 5 weights instead of 4. Let's make them [0.2, 0.2, 0.2, 0.2, 0.2] or [0.3, 0.3, 0.3].

photosynthesis_metrics/gmsd.py
denproc: The size depends on the number of scales used to calculate the measure. If user would like to use [.1]*10 weights, the size should be at least 2**len(weights) + 1.

zakajd:
Discussed with @denproc that it's better to have support for custom number of scales, than not to have. 
Now user can write `scale_weights = [1., 5., 2.]` and everything will work.

PR #78, opened 2020-06-03
Update TV interface

### LINKED ISSUE #77: Update TV interface
TV metric and loss were meant to be unified (#76). It meant that function will be removed so it was not updated to work with 2/3/4D tensors.
However, it was decided to not to do that and now TV needs an update.

- Move `validate_input` and `adjust_dimensions` to the `total_variation` function
- Call `total_variation` with all required checks from `TVLoss`

snk4tr:
Closes #77 

## Proposed Changes
- Move validate_input and adjust_dimensions to the total_variation function
- Call total_variation with all required checks from TVLoss
- Adjust tests

### DIFF
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -22,7 +22,9 @@ def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type: 
         https://www.wikiwand.com/en/Total_variation_denoising
         https://remi.flamary.com/demos/proxtv.html
     """
-    assert x.dim() == 4, f'Expected 4D tensor, got {x.size()}'
+    _validate_input(x)
+    x = _adjust_dimensions(x)
+
     if reduction_type == 'l1':
         w_variance = torch.sum(torch.abs(x[:, :, :, 1:] - x[:, :, :, :-1]), dim=[1, 2, 3])
         h_variance = torch.sum(torch.abs(x[:, :, 1:, :] - x[:, :, :-1, :]), dim=[1, 2, 3])
@@ -101,8 +103,6 @@ class TVLoss(_Loss):
         Returns:
             Value of TV loss to be minimized.
         """
-        _validate_input(prediction)
-        prediction = _adjust_dimensions(prediction)
         return self.compute_metric(prediction)
 
     def compute_metric(self, prediction: torch.Tensor) -> torch.Tensor:
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -22,7 +22,7 @@ def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type: 
         https://www.wikiwand.com/en/Total_variation_denoising
         https://remi.flamary.com/demos/proxtv.html
     """
-    _validate_input(x)
+    _validate_input(x, supports_5d=False)
     x = _adjust_dimensions(x)
 
     if reduction_type == 'l1':
--- a/photosynthesis_metrics/tv.py
+++ b/photosynthesis_metrics/tv.py
@@ -22,7 +22,7 @@ def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type: 
         https://www.wikiwand.com/en/Total_variation_denoising
         https://remi.flamary.com/demos/proxtv.html
     """
-    _validate_input(x, supports_5d=False)
+    _validate_input(x, allow_5d=False)
     x = _adjust_dimensions(x)
 
     if reduction_type == 'l1':
--- a/photosynthesis_metrics/utils.py
+++ b/photosynthesis_metrics/utils.py
@@ -24,6 +24,7 @@ def _adjust_dimensions(input_tensors: Union[torch.Tensor, Tuple[torch.Tensor, to
 
 def _validate_input(
         input_tensors: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
+        supports_5d: bool,
         kernel_size: Optional[int] = None,
         scale_weights: Union[Optional[Tuple[float]], Optional[List[float]], Optional[torch.Tensor]] = None) -> None:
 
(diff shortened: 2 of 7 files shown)


photosynthesis_metrics/tv.py
```diff
@@ -22,7 +22,9 @@ def total_variation(x: torch.Tensor, size_average: bool = True, reduction_type:
         https://www.wikiwand.com/en/Total_variation_denoising
         https://remi.flamary.com/demos/proxtv.html
     """
-    assert x.dim() == 4, f'Expected 4D tensor, got {x.size()}'
+    _validate_input(x)
+    x = _adjust_dimensions(x)
```
denproc, line 26: Tensor of `(3, 3, 256, 256, 2)` size will pass that check. SSIM can deal with 2/3/4/5D, while other losses deal only with 2/3/4D tensors. Let's discuss how we can resolve it in a better way than one more `assert`.

snk4tr: We can add a parameter to the `_validate_input` function, which indicates support for 5D. 
The interface for TV will be like:
`_validate_input(x, supports_5d=False)`

snk4tr: @denproc what do you think?

denproc: Looks like a plan 👌🏻. +1

snk4tr:
@denproc done

photosynthesis_metrics/utils.py
```diff
@@ -24,6 +24,7 @@ def _adjust_dimensions(input_tensors: Union[torch.Tensor, Tuple[torch.Tensor, to
 
 def _validate_input(
         input_tensors: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
+        supports_5d: bool,
```
denproc, line 27: What do you think about default value? 5D size is used only in case of ssim, others do not support 5D.
```suggestion
        supports_5d: Optional[bool] = False,
```

denproc, line 27: Does "enable_5d" sound better?

tests/test_utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def test_breaks_if_tensors_have_different_n_dims(tensor_4d: torch.Tensor, tensor_5d: torch.Tensor) -> None:
+    with pytest.raises(AssertionError):
+        _validate_input((tensor_4d, tensor_5d), supports_5d=True)
+
+
+def test_works_if_kernel_size_is_odd(tensor_4d: torch.Tensor) -> None:
+    for kernel_size in [i * 2 + 1 for i in range(2, 42)]:
+        _validate_input(tensor_4d, supports_5d=False, kernel_size=kernel_size)
```
denproc, line 82: try/except pattern?

tests/test_utils.py
```diff
@@ -0,0 +1,104 @@
...
+            _validate_input(inp, supports_5d=False)
+
+
+def test_breaks_if_too_many_tensors_provided(tensor_4d: torch.Tensor) -> None:
+    max_number_of_tensors = 2
+    for n_tensors in range(max_number_of_tensors + 1, (max_number_of_tensors + 1) * 10):
+        inp = tuple(tensor_4d.clone() for _ in range(n_tensors))
+        with pytest.raises(AssertionError):
+            _validate_input(inp, supports_5d=False)
+
```
denproc, line 38: Please, add test to check 2/3/4/5d functionality.

photosynthesis_metrics/utils.py
snk4tr: In general I avoid default values where possible because they enforce error-prone implicit behaviour

snk4tr: Sounds better, but I think that `supports_5d` better reflects the fact that the current metric supports 5d tensors

snk4tr: As discussed offline, the new naming will be `allow_5d` because:
- Reflects action in contrast to `supports` which is more intuitive for functions
- Shows that we **allow** such type of an input in this particular case

snk4tr:
@denproc done

tests/test_utils.py
```diff
@@ -0,0 +1,139 @@
...
+        if tensor.dim() < 5:
+            try:
+                _validate_input(tensor, allow_5d=False)
+            except Exception as e:
+                pytest.fail(f"Unexpected error occurred: {e}")
+        else:
+            with pytest.raises(AssertionError):
+                _validate_input(tensor, allow_5d=False)
+
+    tensor.unsqueeze_(0)
```
denproc, line 67:
```suggestion
    # 1D -> max_num_dims
    max_num_dims = 10
    for _ in range(max_num_dims):
        if 1 < tensor.dim() < 5:
            try:
                _validate_input(tensor, allow_5d=False)
            except Exception as e:
                pytest.fail(f"Unexpected error occurred: {e}")
        else:
            with pytest.raises(AssertionError):
                _validate_input(tensor, allow_5d=False)

        tensor.unsqueeze_(0)
```

tests/test_utils.py
```diff
@@ -0,0 +1,139 @@
...
+        pytest.fail(f"Unexpected error occurred: {e}")
+
+
+def test_breaks_if_5d_tensor_has_wrong_format(tensor_5d_broken: torch.Tensor) -> None:
+    with pytest.raises(Exception):
+        _validate_input(tensor_5d_broken, allow_5d=True)
+
+
+def test_works_on_two_not_5d_tensors(tensor_2d: torch.Tensor) -> None:
+    tensor = tensor_2d.clone()
```
denproc, line 83:
```suggestion
def test_works_on_two_not_5d_tensors(tensor_1d: torch.Tensor) -> None:
    tensor = tensor_1d.clone()
```

tests/test_utils.py
```diff
@@ -0,0 +1,139 @@
...
+    max_num_dims = 10
+    for _ in range(max_num_dims):
+        another_tensor = tensor.clone()
+        if tensor.dim() < 5:
```
denproc, line 87:
```suggestion
        if 1 < tensor.dim() < 5:
```

PR #85, opened 2020-06-11
Add __version__ attribute 

### LINKED ISSUE #81: Add package version to photosynthesis_metrics.__version__
It will be great to have this commonly used attribute to identify the package version

snk4tr:
Closes #81 

## Proposed Changes

  - Add the commonly used attribute to identify the package version

Better be merged after #83

### DIFF
--- a/photosynthesis_metrics/__init__.py
+++ b/photosynthesis_metrics/__init__.py
@@ -1,3 +1,5 @@
+__version__ = "0.4.0"
+
 from .ssim import ssim, multi_scale_ssim, SSIMLoss, MultiScaleSSIMLoss
 from .msid import MSID
 from .fid import FID
@@ -9,6 +11,3 @@ from .isc import IS, inception_score
 from .vif import VIFLoss, vif_p
 from .brisque import BRISQUELoss, brisque
 from .psnr import psnr
-
-
-__version__ = "0.4.0"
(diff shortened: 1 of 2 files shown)


snk4tr:
@denproc let's merge this if there are no additional comments

photosynthesis_metrics/__init__.py
```diff
@@ -8,3 +8,6 @@
 from .isc import IS, inception_score
 from .vif import VIFLoss, vif_p
 from .brisque import BRISQUELoss, brisque
+
+
+__version__ = "0.4.0"
```
denproc, line 13: According to [PEP8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names) the `__version__` should be defined before import statements.

setup.py
```diff
@@ -1,14 +1,33 @@
+import os
 import setuptools
```
denproc, line 3: I would suggest to use `from photosynthesis_metrics import __version__` as we do not use additional files with metadata.

photosynthesis_metrics/__init__.py
snk4tr: Hmm it means that [even pip does it the wrong way](https://github.com/pypa/pip/blob/master/src/pip/__init__.py) 😄
Ok, I will change that

setup.py
snk4tr: Discussed with @denproc offline and decided to leave the existing implementation since it is more conventional and robust one

snk4tr:
@denproc ready

PR #87, opened 2020-06-11
Fix bug in BRISQUE not supporting GPU

zakajd:
#  Proposed Changes
- New tests to make sure metric works on GPU
- Small fixes in tensor initialisation to correctly infer device type

### DIFF
--- a/tests/test_brisque.py
+++ b/tests/test_brisque.py
@@ -23,8 +23,26 @@ def test_brisque_if_works_with_grey(prediction_grey: torch.Tensor) -> None:
         pytest.fail(f"Unexpected error occurred: {e}")
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason='No need to run test if there is no GPU.')
+def test_brisque_if_works_with_grey_on_gpu(prediction_grey: torch.Tensor) -> None:
+    try:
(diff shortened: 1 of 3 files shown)


tests/test_brisque.py
```diff
@@ -23,13 +23,31 @@ def test_brisque_if_works_with_grey(prediction_grey: torch.Tensor) -> None:
         pytest.fail(f"Unexpected error occurred: {e}")
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason='No need to run test if there is no GPU.')
+def test_brisque_if_works_with_grey_on_gpu(prediction_grey: torch.Tensor) -> None:
+    try:
```
PavelOstyakov, line 28: Hi, just a question related to your tests writing style. Why do you use:
```
try:
   do_someting()
except Exception as e:
   pytest.fail(f"Unexpected error occurred: {e}")
```

Instead of just:
```
do_something()
```

pytest will automatically handle any exception and print its message, as well as the whole traceback.

zakajd: No particular reason. 
I just copy/pasted this style from some other repo  🤷‍♂️
Will reconsider this practise after your comment

PR #88, opened 2020-06-11
Bug fix for BRISQUE score

### LINKED ISSUE #86: BRISQUE returns nan
Hi, I've found that if a tensor contains all (or almost all) zeros, brisque returns nan. Is this behavior expected?

```
In [1]: from photosynthesis_metrics import brisque

In [2]: import torch

In [3]: prediction = torch.zeros(3, 3, 256, 256)

In [4]: brisque(prediction)
Out[4]: tensor(nan)

In [5]: prediction[:, :, 0, 0] = 1

In [6]: brisque(prediction)
Out[6]: tensor(nan)
```

denproc:
Closes #86

## Proposed Changes
  - Improved robustness for input tensors with small variance of values

### DIFF
--- a/photosynthesis_metrics/brisque.py
+++ b/photosynthesis_metrics/brisque.py
@@ -22,9 +22,12 @@ def _ggd_parameters(x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
 
     sigma_sq = x.pow(2).mean(dim=(-1, -2))
     sigma = sigma_sq.sqrt().squeeze(dim=-1)
+
+    assert not torch.isclose(sigma, torch.zeros_like(sigma)).all(), \
+        'Expected image with non zero variance of pixel values'
+
     E = x.abs().mean(dim=(-1, -2))
-    EPS = torch.finfo(torch.float32).eps
-    rho = sigma_sq / (E ** 2 + EPS)
+    rho = sigma_sq / E ** 2
 
     indexes = (rho - r_table).abs().argmin(dim=-1)
     solution = gamma[indexes]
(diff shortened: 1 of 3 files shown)


photosynthesis_metrics/brisque.py
```diff
@@ -32,16 +36,25 @@ def _ggd_parameters(x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
...
+    r_table = r_table.repeat(x.size(0), 1)
 
     mask_left = x < 0
     mask_right = x > 0
-    count_left = mask_left.sum(dim=(-1, -2))
-    count_right = mask_right.sum(dim=(-1, -2))
+    count_left = mask_left.sum(dim=(-1, -2), dtype=torch.float32)
+    count_right = mask_right.sum(dim=(-1, -2), dtype=torch.float32)
+
+    assert (count_left > 0).all(), 'Expected input tensor (pairwise products of neighboring MSCN coefficients)' \
```
PavelOstyakov, line 47: This is not what a typical user expects from a metric. Currently, I should check all images manually before passing to a metric. Since I'm gonna use it during GANs training, sometimes images can not meet this criterion. Why not add EPS? In this case, I won't have to worry about it.

denproc: I agree that this is not a typical experience. In the first commit I added EPS to avoid nan. However, after closer look at BRISQUE theory, I decided to add asserts for now. This results from the estimation of parameters of generalised gaussian distribution where \sigma should be strictly greater zero. In case of images with \sigma = 0, we deal with Dirac function instead of GGD. Same applies to AGGD. 

In addition available implementations return NaN as well. 
For now, I would suggest to assert such inputs in accordance to theory instead of returning undefined finite number.

snk4tr: @PavelOstyakov I agree with @denproc that undefined finite number in case of a zero-filled tensor input does not make much sense. 
I would also argue that zero-filled tensor is quite a unique thing in a real setting in general and in GAN training in particular. May be we do not see something and you could provide a bit more context on why the behavior that you propose is strictly better than the one which @denproc proposed?

PR #95, opened 2020-06-12
Add PSNR metric

### LINKED ISSUE #93: Add PSNR metric
I know it's easy to implement it yourself, but it's better to have it with all other metrics

Implement basic PSNR metric

zakajd:
Closes #93 

## Proposed Changes

  - PSNR metric
  - Tests
  - Updated README.md

### DIFF
--- a/photosynthesis_metrics/psnr.py
+++ b/photosynthesis_metrics/psnr.py
@@ -21,7 +21,7 @@ def psnr(prediction: torch.Tensor, target: torch.Tensor,
         PSNR: Index of similarity betwen two images.
     Note:
         Implementaition is based on Wikepedia https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio
-        Colour images a first converted to YCbCr format and only luminance component is considered.
+        Colour images are first converted to YCbCr format and only luminance component is considered.
     """
     _validate_input((prediction, target), allow_5d=False)
 
--- a/photosynthesis_metrics/psnr.py
+++ b/photosynthesis_metrics/psnr.py
@@ -1,0 +1,55 @@
+r""" This module implements Peak Signal-to-Noise Ration (PSNR) in PyTorch.
+"""
+import torch
+from typing import Optional, Union
+
+from photosynthesis_metrics.utils import _validate_input
+
+
+def psnr(prediction: torch.Tensor, target: torch.Tensor,
+         data_range: Union[int, float] = 1.0, reduction: Optional[str] = 'mean'):
+    r"""Compute Peak Signal-to-Noise Ration for a batch of images.
+    Supports both greyscale and color images with RGB channel order.
+
+    Args:
+        prediction: Batch of predicted images with shape (batch_size x channels x H x W)
+        target: Batch of target images with shape  (batch_size x channels x H x W)
+        data_range: Value range of input images (usually 1.0 or 255). Default: 1.0
+        reduction: Reduction over samples in batch: "mean"|"sum"|"none"
+        
+    Returns:
+        PSNR: Index of similarity betwen two images.
+    Note:
+        Implementaition is based on Wikepedia https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio
+        Colour images a first converted to YCbCr format and only luminance component is considered.
+    """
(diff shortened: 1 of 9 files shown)


photosynthesis_metrics/psnr.py
```diff
@@ -0,0 +1,55 @@
...
+    psnr = - 10 * torch.log10(mse + EPS)
+
+    if reduction == 'mean':
+        return psnr.mean(dim=0)
+    elif reduction == 'sum':
+        return psnr.sum(dim=0)
+    elif reduction == 'none':
+        return psnr
+    else:
+        raise ValueError(f'Expected reduction modes are "mean"|"sum"|"none", got {reduction}')
```
snk4tr, line 55: Consider reimplmeneting it [the way it is done in BRISQUE](https://github.com/photosynthesis-team/photosynthesis.metrics/blob/83e65bb48e0e61b6b89ae75aaa3fcbb63b74dac5/photosynthesis_metrics/brisque.py#L178)

tests/test_psnr.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture(scope='module')
+def prediction() -> torch.Tensor:
+    return torch.rand(4, 3, 256, 256)
+
+
+@pytest.fixture(scope='module')
+def target() -> torch.Tensor:
+    return torch.rand(4, 3, 256, 256)
```
snk4tr, line 15: Please check with at least several sized on tensors.
Also check edge cases to see what happens when `torch.zeros()`, `torch.ones()`, empty tensors and things like that are passed

photosynthesis_metrics/psnr.py
```diff
@@ -0,0 +1,55 @@
...
+    _validate_input((prediction, target), allow_5d=False)
+
+    # Constant for numerical stability
+    EPS = 1e-8
+
+    if data_range == 255:
+        prediction = prediction / 255.
+        target = target / 255.
+
+    # Convert RGB image to YCbCr and take luminance: Y = 0.299 R + 0.587 G + 0.114 B
```
PavelOstyakov, line 35: It is not a common way to compute PSNR. Usually, PSNR on RGB images means -10 log(mse) where mse is computed across all three channels.

zakajd: PSNR is historically defined only for greyscale images. There is no consensus how to do it for RGB, I've seen both approaches (compute on luminance and compute on each channel separately). 
I'll add a flag to compute it on all 3 channels

PavelOstyakov: If you open any repo with RGB image enhancement problem (SR, Denoising, etc..) or any challenge (like NTIRE, AIM, etc.), in 99.99% cases PSNR will be computed on 3 channels. So, this definitely should be the default behavior.

zakajd: Added flag `convert_to_greyscale` which is `False` by default.
Now results match other implementations

photosynthesis_metrics/psnr.py
```diff
@@ -0,0 +1,53 @@
...
+    y = y / data_range
+
+    if (x.size(1) == 3) and convert_to_greyscale:
+        # Convert RGB image to YCbCr and take luminance: Y = 0.299 R + 0.587 G + 0.114 B
+        x = 0.299 * x[:, 0, :, :] + 0.587 * x[:, 1, :, :] + 0.114 * x[:, 2, :, :]
+        y = 0.299 * y[:, 0, :, :] + 0.587 * y[:, 1, :, :] + 0.114 * y[:, 2, :, :]
+
+        # Add channel dimension
+        x = x[:, None, :, :]
+        y = y[:, None, :, :]
```
denproc, line 43: Consider this approach to avoid slicing.
`if x.size(1) == 3:
        # rgb_to_grey - weights to transform RGB image to grey
        rgb_to_grey = torch.tensor([0.299, 0.587, 0.114]).view(1, -1, 1, 1).to(x)
        x = torch.sum(x * rgb_to_grey, dim=1, keepdim=True)
`

zakajd: Good point. Added this

Issue #54, opened 2020-05-20
Feature: LPIPS metric/loss


zakajd:
**Is your feature request related to a problem? Please describe.**
Add [LPIPS measure](http://arxiv.org/abs/1801.03924), which is somewhat similar to Perceptual score.  
Differences:
- Predictions are made on image patches
- Weights for different feature maps are learnable
- Feature maps are normalized to have same scale.

**Describe the solution you'd like**
Adjust code from official implementation repo: https://github.com/richzhang/PerceptualSimilarity

This should be closed before #26 , because then Perceptual score can be inherited from this score with some parameters fixed.

PR #96, opened 2020-06-15
feature/issue#26: Perceptual losses

### LINKED ISSUE #26: Measure: Perceptual Scores
Need to add two types of perceptual scores: content score and style score. Both scores might be needed both and measures and as losses.

Add both measures based on implementation by @zakajd. Improvements to be made:
- Readable layer names. They are needed for better understanding of what layers are used.
- Normalization of values in layers if several layers are used.

zakajd:
Closes #26 and  #54 

## Proposed Changes
  - Content loss
  - Style loss
- LPIPS

- Feature normalization to unit length
- Optional VGG16 or VGG19 extractor
- Any number of layers supported
- Any weights supported. Can be per-level (Style transfer case), or per-feature (as in LPIPS)


zakajd:
Ready to be reviewed.
I'll add more tests later, for now please, check that there no big issues with class functions.

I added weights for LPIPS into latest release, so they can be downloaded the same way, as BRISQUE weights

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,283 @@
...
+        r"""Compute Gram matrix for batch of features.
+        Args:
+            x: Tensor of shape BxCxHxW
+        """
+        B, C, H, W = x.size()
+        gram = []
+        for i in range(B):
+            features = x[i].view(C, H * W)
+            
+            # Normalize gram matrix by deviding into number of elements
```
zakajd, line 241:
```suggestion
            
```

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,283 @@
+"""
+Implementation of VGG16 loss, originaly used for style transfer and usefull in many other task (including GAN training)
+It's work in progress, no guarantees that code will work
```
zakajd, line 3:
```suggestion
```

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+    "conv5_4": 34, "relu5_4": 35,
+    "pool5": 36,
+}
+
+
+class ContentLoss(_Loss):
+    r"""Creates Content loss that can be used for image style transfer of as a measure in
+    image to image tasks.
+    Uses pretrained VGG models from torchvision. Normalizes features before summation.
+    Expects input to be in range [0, 1] or normalized with ImageNet statistics into range [-1, 1]
```
denproc, line 65: Range [-1, 1] is invalid according to checks in `_validate_input`.
Same for following losses.

tests/test_perseptual.py
```diff
@@ -0,0 +1,55 @@
...
+
+# ================== Test class: `ContentLoss` ==================
+def test_content_loss_init(prediction: torch.Tensor, target: torch.Tensor) -> None:
+    ContentLoss()
+
+
+def test_content_loss_forward(prediction: torch.Tensor, target: torch.Tensor, device: str) -> None:
+    loss = ContentLoss()
+    loss(prediction.to(device), target.to(device))
+
```
denproc, line 36: Let's check `backward()` for each loss.
Same for the following losses.

tests/test_perseptual.py
```diff
@@ -0,0 +1,55 @@
...
+    devices.append("cuda")
+
+
+@pytest.fixture(scope='module', params=devices)
+def device(request: Any) -> Any:
+    return request.param
+
+
+# ================== Test class: `ContentLoss` ==================
+def test_content_loss_init(prediction: torch.Tensor, target: torch.Tensor) -> None:
```
denproc, line 29: Do we need such test as the initialisation is checked in forward test?
If so, the arguments are not needed.
Same for the following losses.

tests/test_perseptual.py
```diff
@@ -0,0 +1,55 @@
...
+
+devices: List[str] = ["cpu"]
+if torch.cuda.is_available():
+    devices.append("cuda")
+
+
+@pytest.fixture(scope='module', params=devices)
+def device(request: Any) -> Any:
+    return request.param
+
```
denproc, line 26: Let's add tests with zeros/ones/minus one/other extreme values?

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+                    self.model.features[i] = torch.nn.AvgPool2d(kernel_size=2, stride=2, padding=0)
+
+        # Disable gradients
+        for param in self.model.parameters():
+            param.requires_grad_(False)
+
+        self.distance = {
+            "mse": nn.MSELoss(reduction='none'),
+            "mae": nn.L1Loss(reduction='none'),
+        }[distance]
```
denproc, line 121:
```suggestion
        self.distance = {
            "mse": nn.MSELoss,
            "mae": nn.L1Loss,
        }[distance](reduction='none')
```

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
+"""
+Implementation of VGG16 loss, originaly used for style transfer and usefull in many other task (including GAN training)
+"""
```
snk4tr, line 3: Since we have references to initial papers implementations in other metrics, it would be great to also add it here

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    # Constant used in feature normalization to avoid zero division
+    EPS = 1e-10
+
+    def __init__(self, feature_extractor: str = "vgg16", layers: List[str] = ["relu3_3"],
+                 weights: List[Union[float, torch.Tensor]] = [1.], replace_pooling: bool = False, distance: str = "mse",
+                 reduction: str = "mean", normalize_input: bool = True, normalize_features: bool = False) -> None:
+        r"""
+        Args:
+            feature_extractor: Name of model used to extract features. One of {`vgg16`, `vgg19`}
```
snk4tr, line 75: Can we support addition of external feature extractors? 
I think it will be useful when we get to the question of how domain on which feature extractor was trained influences the quality of features.

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+            layers: List of string with layer names. Default: [`relu3_3`]
+            weights: List of float weight to balance different layers
+            replace_pooling: Flag to replace MaxPooling layer with AveragePooling. See [1] for details.
```
snk4tr, line 78: Would `use_average_pooling` set to `False` by default be a better naming here?

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+            param.requires_grad_(False)
+
+        self.distance = {
+            "mse": nn.MSELoss(reduction='none'),
+            "mae": nn.L1Loss(reduction='none'),
+        }[distance]
+
+        self.weights = [torch.tensor(w) for w in weights]
+        mean = torch.tensor([0.485, 0.456, 0.406]) if normalize_input else torch.tensor([0., 0., 0.])
+        std = torch.tensor([0.229, 0.224, 0.225]) if normalize_input else torch.tensor([1., 1., 1.])
```
snk4tr, line 125: Please move `torch.tensor([0.485, 0.456, 0.406])` and `torch.tensor([0.229, 0.224, 0.225])` to a separate variable with proper names which describe where these values are taken from

photosynthesis_metrics/perceptual.py
snk4tr: Btw it is interesting why tests did not fail because of that. I think it shows that tests simply do not cover this case, which has to be fixed

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+        prediction_features = self.get_features(prediction)
+        target_features = self.get_features(target)
+        
+        distances = [self.distance(x, y) for x, y in zip(prediction_features, target_features)]
+
+        # Scale distances, then average in spatial dimensions, then stack and sum in channels dimension
+        loss = torch.cat([(d * w.to(d)).mean(dim=[2, 3]) for d, w in zip(distances, self.weights)], dim=1).sum(dim=1)
+
+        # Solve big memory consumption
+        torch.cuda.empty_cache()
```
snk4tr, line 160: Be careful with that. It may result in a huge performance degradation because GPU I/O frequently takes way more time then computations. 
Also note that this code actually effects the user of the library who may use this metric during training/inference. In this case it may lead not only in slow computation of the metric, but also performance degradation in the whole training/validation pipeline.

I would remove it.

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+
+        if self.reduction == 'none':
+            return loss
+
+        return {'mean': loss.mean,
+                'sum': loss.sum
+                }[self.reduction](dim=0)
```
snk4tr, line 167: Probably its time to extract this code that is getting copy-pasted in all metrics that support reduction

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        norm_factor = torch.sqrt(torch.sum(x ** 2, dim=1, keepdim=True))
+        return x / (norm_factor + self.EPS)
+
+
+class StyleLoss(ContentLoss):
+    r"""Creates Style loss that can be used for image style transfer or as a measure in
+    image to image tasks.
+    Uses pretrained VGG models from torchvision. Features can be normalized before summation.
+    Expects input to be in range [0, 1] or normalized with ImageNet statistics into range [-1, 1]
```
snk4tr, line 199: I think it is worth mentioning here that this is the loss that uses Gramm matrix computions

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+        self.distance = {
+            "mse": nn.MSELoss(reduction='none'),
+            "mae": nn.L1Loss(reduction='none'),
+        }[distance]
+
+        self.weights = [torch.tensor(w) for w in weights]
+        mean = torch.tensor([0.485, 0.456, 0.406]) if normalize_input else torch.tensor([0., 0., 0.])
+        std = torch.tensor([0.229, 0.224, 0.225]) if normalize_input else torch.tensor([1., 1., 1.])
+        self.mean = mean.view(1, 3, 1, 1)
+        self.std = std.view(1, 3, 1, 1)
```
snk4tr, line 127: I would also make these to be parameters. 
Suppose you want to standardize your medical imaging dataset. These values will not work

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+        prediction_gram = [self.gram_matrix(x) for x in prediction_features]
+        target_gram = [self.gram_matrix(x) for x in target_features]
+
+        distances = [self.distance(x, y) for x, y in zip(prediction_gram, target_gram)]
+
+        # Scale distances, then average in spatial dimensions, then stack and sum in channels dimension
+        loss = torch.stack([(d * w.to(d)).mean(dim=[1, 2]) for d, w in zip(distances, self.weights)], dim=1).sum(dim=1)
+
+        # Solve big memory consumption
+        torch.cuda.empty_cache()
```
snk4tr, line 221: Same as above

photosynthesis_metrics/perceptual.py
```diff
@@ -0,0 +1,280 @@
...
+
+        if self.reduction == 'none':
+            return loss
+
+        return {'mean': loss.mean,
+                'sum': loss.sum
+                }[self.reduction](dim=0)
```
snk4tr, line 228: Same as above

tests/test_perseptual.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.fixture(scope='module')
+def prediction() -> torch.Tensor:
+    return torch.rand(4, 3, 128, 128)
+
+
+@pytest.fixture(scope='module')
+def target() -> torch.Tensor:
+    return torch.rand(4, 3, 128, 128)
```
snk4tr, line 15: It's time to extend tests. Do not test your code on only one ideal input data option

piq/perceptual.py
```diff
@@ -0,0 +1,304 @@
...
+    
+    def compute_metric(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        # Normalize input
+        mean, std = self.mean.to(prediction), self.std.to(prediction)
+        prediction = (prediction - mean) / std
+        target = (target - mean) / std
+
+        self.model.to(prediction)
+        prediction_features = self.get_features(prediction)
+        target_features = self.get_features(target)
```
snk4tr, line 170: This code is duplicated -lines 230 - 236. 
Replace it with a new fuction/method or include this functionality in `self.get_features`.

piq/perceptual.py
```diff
@@ -0,0 +1,304 @@
...
+    def max_pool_to_average_pool(self, module: torch.nn.Module) -> torch.nn.Module:
+        r"""Turn All MaxPool layers into AveragePool"""
+        module_output = module
+        if isinstance(module, torch.nn.MaxPool2d):
+            module_output = torch.nn.AvgPool2d(kernel_size=2, stride=2, padding=0)
+            
+        for name, child in module.named_children():
+            module_output.add_module(name, self.max_pool_to_average_pool(child))
+        del module
+        return module_output
```
snk4tr, line 218: I don't understand why you first copy the reference to the `module` object, then modify it and then delete the initial reference in case if you reassign the output to the same object (`self.model = self.max_pool_to_average_pool(self.model)`, line 130) anyway.
To me it seems like deleting lines 211 and 217 changes nothing. Can you comment on that?

piq/perceptual.py
zakajd: Unified interface for both classes by moving most of operations into `forward` call

piq/perceptual.py
zakajd: You are right, `del` isn't necessary. 
Without copy it fails when tries to call `add_module` for not existing `module_output` variable.

tests/test_perseptual.py
zakajd: Done 👍

tests/test_perseptual.py
zakajd: Done 👍

tests/test_perseptual.py
zakajd: Added tests on bigger number. of input tensors

PR #103, opened 2020-06-17
Bugfix for FID

### LINKED ISSUE #100: FID doesn't always support cuda
**Describe the bug**
If in FID `covmean`is infinity, small offset is added before computation of matrix square root.
[This line](https://github.com/photosynthesis-team/photosynthesis.metrics/blob/c25329c0fea0177559c084b9c64fa85916faf1e9/photosynthesis_metrics/fid.py#L86) doesn't handle case of CUDA tensors properly and always creates `offset` tensor on CPU. Such edge case isn't tested, thats why this bug went unnoticed.
**To Reproduce**
```
prediction_features = torch.rand(4, 128).cuda()
target_features = torch.zeros(4, 128).cuda()
loss = FID()
loss(prediction_features, target_features)
```
**Expected behavior**
Suggested fix to support CUDA tensors:
`offset = torch.eye(sigma1.size(0)).to(mu1) * eps`

**Screenshots**
image

**Additional context**

### LINKED ISSUE #101: Strange FID behaviour on CPU
**Describe the bug**
When computations are done on CPU multiple times for the same features, after first run we get different results

**To Reproduce**
```
prediction_features = torch.rand(4, 128)
target_features = torch.zeros(4, 128)

loss = FID()
for i in range(3):
    print(loss(prediction_features, target_features))
```

**Expected behavior**
Value of FID should be the same all the times.

**Screenshots**
image

**Additional context**
There is no such problem on GPU, it only happens for CPU tensors

### LINKED ISSUE #102: FID returns NaN
**Describe the bug**
For some combination of input and target features FID computation returns NaN

**To Reproduce**
1. [Download `prediction_target_features.pt`](https://yadi.sk/d/CNd9Wy8eV9_OdA)
2. Run the code:
```
data = torch.load('prediction_target_features.pt')
prediction_features = data[0]
target_features = data[1]
loss = FID()
# CUDA fixed with `offset = torch.eye(sigma1.size(0)).to(mu1) * eps`
loss(prediction_features.cuda(), target_features.cuda())
```
3. See NaN value
`tensor(nan, device='cuda:0')`

**Expected behavior**
Error message or any meaningful number 

**Additional context**
May be related to #101

denproc:
Closes #100, Closes #101, Closes #102 

## Proposed Changes

  - Removed inlace operation, which changed input tensors
  - Transferred tensors to the type and device of the input
  - Removed unused variable
  - Changed float precision to `float64` to avoid related errors in computation (https://github.com/pytorch/pytorch/issues/20551)

### DIFF
--- a/photosynthesis_metrics/fid.py
+++ b/photosynthesis_metrics/fid.py
@@ -38,7 +38,6 @@ def _sqrtm_newton_schulz(A: torch.Tensor, num_iters: int = 100) -> Tuple[torch.T
 
     if num_iters <= 0:
         raise ValueError(f'Number of iteration equals {num_iters}, expected greater than 0')
-    dtype = A.type()
     dim = A.size(0)
     normA = A.norm(p='fro')
     Y = A.div(normA)
--- a/photosynthesis_metrics/fid.py
+++ b/photosynthesis_metrics/fid.py
@@ -82,7 +82,7 @@ def _compute_fid(mu1: torch.Tensor, sigma1: torch.Tensor, mu2: torch.Tensor, sig
     # Product might be almost singular
     if not torch.isfinite(covmean).all():
         print(f'FID calculation produces singular product; adding {eps} to diagonal of cov estimates')
-        offset = torch.eye(sigma1.size(0)) * eps
+        offset = torch.eye(sigma1.size(0)).to(mu1) * eps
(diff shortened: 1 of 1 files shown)


denproc:
@zakajd , ready for review

PR #110, opened 2020-06-23
BRISQUE: Data ranges.

### LINKED ISSUE #108: BRISQUE metric doesn't work
**Describe the bug**
I experimented with BRISQUE score today and found it to have nearly 0 correlation with Mean Opinion Scores for TID2013 dataset.
I compared result with `pybrisque` lib and found that they are very different for real images.

**To Reproduce**
1. Install `pybrisque`
2. Compute value for 2 attached images 
3. Result will be different.

`goldhill.gif`: **27.8** pybrisque / **111.07** piq.brisque
`I01.BMP`: **42.1** pybrisque / **95.5** piq.brisque

**Expected behavior**
I don't know correct value for first image, but for the second it definitely must be ~42, see plot with MATLAB comparison in[ pybrisque repo](https://github.com/bukalapak/pybrisque)

[images.zip](https://github.com/photosynthesis-team/piq/files/4813243/images.zip)

denproc:
Closes #108 

## Proposed Changes

  - Changed data range which is used for estimation of BRISQUE score;
  - Changed tests in accordance to the proper data range.

### DIFF
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -173,7 +173,7 @@ def brisque(x: torch.Tensor,
     _validate_input(input_tensors=x, allow_5d=False)
     x = _adjust_dimensions(input_tensors=x)
 
-    assert data_range >= x.max(), 'Expected data range greater or equal maximum value.'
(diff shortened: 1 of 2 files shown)


denproc:
@zakajd, ready for review as tests are passed

PR #116, opened 2020-06-25
Test coverate info

snk4tr:
Closes #84 

## Proposed Changes

  - Add test coverage info to the testing CI pipeline
  - Integration with [codecov](https://codecov.io/) via GitHub Actions
  - Display of coverage info with a badge in the README

PR #119, opened 2020-06-30
Add Visual Saliency-induced Index (VSI)

### LINKED ISSUE #115: Metric: VSI (Visual Saliency-induced Index)
According to multiple studies VSI shows very good performance on broad set of image domains. It's mentioned in almost all IQA papers after 2016 as one of the best methods. There is no Python implementation and original MATLAB code doesn't support batch computation.

Rewrite MATLAB -> PyTorch and add batching support.
Optionally add support for gradient optimization
**Additional context**
Original paper: https://ieeexplore.ieee.org/document/6873260
MATLAB code from authors: http://sse.tongji.edu.cn/linzhang/IQA/VSI/VSI.htm

denproc:
Closes #115 

## Proposed Change
- Added VSI measure
- Added VSI Loss function
- Both support back propagation
- Both support GPU
- Tests to check functionality

Let's merge #116 first.


zakajd:
Nice job!
Have you compared results from your implementation with MATLAB?
Just for test I run your code on 2 images from TID2013 dataset (they are attached in zip archive) and got score `0.9521`

Code that I used
```
import torch
from skimage.io import imread
from piq import vsi

I01 = torch.tensor(imread('data/I01.BMP')).permute(2, 0, 1)
i1_01_5 = torch.tensor(imread('data/i01_01_5.bmp')).permute(2, 0, 1)
print(vsi(i1_01_5, I01, data_range=255.))
```

While computing VSI score for those 2 images in OCTAVE (I don't have MATLAB installed) produces score `0.96405`.
I'm using [octave-online](https://octave-online.net), after loading `VSI.m` file and 2 images code to compute score is simple:

```
I01 = imread("I01.BMP")
i01_01_5 = imread("i01_01_5.bmp")
score = VSI(i01_01_5, I01)
```

Difference is quite significant. Can you comment on that?

[test_images.zip](https://github.com/photosynthesis-team/piq/files/4859232/test_images.zip)

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+    vs_prediction = avg_pool2d(vs_prediction.unsqueeze(1), kernel_size=kernel_size).squeeze(1)
+    vs_target = avg_pool2d(vs_target.unsqueeze(1), kernel_size=kernel_size).squeeze(1)
+
+    prediction_LMN = avg_pool2d(prediction_LMN, kernel_size=kernel_size)
+    target_LMN = avg_pool2d(target_LMN, kernel_size=kernel_size)
+
+    gm_prediction = scharr_grad_map(prediction_LMN[:, 1].unsqueeze(1)).squeeze(1)
+    gm_target = scharr_grad_map(target_LMN[:, 1].unsqueeze(1)).squeeze(1)
+
+    s_vs = (2 * vs_prediction * vs_target + c1) / (vs_prediction.pow(2) + vs_target.pow(2) + c1)
```
zakajd, line 73: This type of similarity computation is also very common. SSIM, MS-SSIM, GMSD, VIF and many other not yet implemented metrics use it
How about doing something like this:
```
def similarity_map(map1, map2, constant):
     return (2 * map1 * map2 + constant) / (map2.pow(2) + map2.pow(2) + constant)
```

```suggestion
    s_vs = similarity_map(vs_prediction, vs_target, c1)
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+
+    r = (xx.pow(2) + yy.pow(2)).sqrt()
+    r[0, 0] = 1
+
+    lg = torch.exp((-(r / omega_0).log().pow(2)) / (2 * sigma_f ** 2))
+    lg[0, 0] = 0
+    return lg
+
+
+def ifftshift(x: torch.Tensor):
```
zakajd, line 283: `ifftshift` is also used in FSIM. Let's move into separate file.

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+
+    x_lab = torch.matmul(tmp.permute(0, 2, 3, 1), weights_XYZ_to_LAB.t()).permute(0, 3, 1, 2) + bias_XYZ_to_LAB
+    return x_lab
+
+
+def rgb2lab(x: torch.Tensor, data_range: float = 255.) -> torch.Tensor:
+    return xyz2lab(rgb2xyz(x / data_range))
+
+
+def log_gabor(size: tuple, omega_0: float = 0.021, sigma_f: float = 1.34) -> torch.Tensor:
```
zakajd, line 264: Add `from typing import Tuple` on top

```suggestion
def log_gabor(size: Tuple[int, int], omega_0: float = 0.021, sigma_f: float = 1.34) -> torch.Tensor:
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+    xx, yy = torch.meshgrid((torch.arange(0, size[0], dtype=torch.float32) - size[0] // 2) / (size[0] - size[0] % 2),
+                            (torch.arange(0, size[1], dtype=torch.float32) - size[1] // 2) / (size[1] - size[1] % 2))
+
+    mask = xx.pow(2) + yy.pow(2) <= 0.25
```
zakajd, line 268: In PyTroch `.pow(2)` and `** 2` are equivalent, but I would argue that the second format is easier to read and understand.

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
+r"""Implemetation of Visual Saliency-induced Index
+Code is based on MATLAB version for computations in pixel domain
```
snk4tr, line 2:
```suggestion
Code is based on the MATLAB version for computations in pixel domain
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+from torch.nn.functional import conv2d, avg_pool2d, interpolate
+from .utils import _validate_input, _adjust_dimensions
+
+
+def vsi(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean', data_range: float = 1.,
+        c1: float = 1.27, c2: float = 386., c3: float = 130., alpha: float = 0.4, beta: float = 0.02,
+        omega_0: float = 0.021, sigma_f: float = 1.34, sigma_d: float = 145., sigma_c: float = 0.001) -> torch.Tensor:
+    r"""Compute Visual Saliency-induced Index for a batch of images.
+
+        Both inputs supposed to have RGB order.
```
snk4tr, line 20:
```suggestion
        Both inputs are supposed to have RGB order.
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+    https://ieeexplore.ieee.org/document/6873260
+"""
+
+import torch
+from torch.nn.modules.loss import _Loss
+from torch.nn.functional import conv2d, avg_pool2d, interpolate
+from .utils import _validate_input, _adjust_dimensions
+
+
+def vsi(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean', data_range: float = 1.,
```
snk4tr, line 15: Since `data_range` can be either `int` or `float`, correct data type will be:
```suggestion
def vsi(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean', data_range: Union[int, float] = 1.,
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Returns:
+            VSI: Index of similarity between two images. Usually in [0, 1] interval.
+        Note:
+            See https://ieeexplore.ieee.org/document/6873260 for details.
+        """
+
+    _validate_input(input_tensors=(prediction, target), allow_5d=False)
+    prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
+    assert prediction.size(-3) == 3 and target.size(-3) == 3, 'Expected RGB images, got images with ' \
```
snk4tr, line 44: Is it possible to adjust the implementation to support grayscale images?

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+    if reduction == 'none':
+        return output
+    return {'mean': torch.mean,
+            'sum': torch.sum
+            }[reduction](output, dim=0)
+
+
+class VSILoss(_Loss):
+    def __init__(self, reduction: str = 'mean', c1: float = 1.27, c2: float = 386., c3: float = 130.,
+                 alpha: float = 0.4, beta: float = 0.02, data_range: float = 1.,
```
snk4tr, line 98: Same as above:
```suggestion
                 alpha: float = 0.4, beta: float = 0.02, data_range: Union[int, float] = 1.,
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+                alpha: power for gradient component of VSI
+                beta: power for color component of VSI
+                omega_0: coefficient to get log Gabor filter at SDSP
+                sigma_f: coefficient to get log Gabor filter at SDSP
+                sigma_d: coefficient to get SDSP
+                sigma_c: coefficient to get SDSP
+
+            Shape:
+                - Input: Required to be 2D (H, W), 3D (C,H,W), 4D (N,C,H,W), channels first.
+                - Target: Required to be 2D (H, W), 3D (C,H,W), 4D (N,C,H,W), channels first.
```
snk4tr, line 123: This is a great way to describe input requirements. I am curios why the same is not present in the `vsi` function 🤔

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+
+    vs_m = interpolate((s_f * s_d * s_c).unsqueeze(1), size[-2:], mode='bilinear', align_corners=True).squeeze(1)
+
+    min_vs_m = vs_m.min(dim=-1, keepdim=True).values.min(dim=-2, keepdim=True).values
+    max_vs_m = vs_m.max(dim=-1, keepdim=True).values.max(dim=-2, keepdim=True).values
+    return (vs_m - min_vs_m) / (max_vs_m - min_vs_m + eps)
+
+
+def rgb2xyz(x: torch.Tensor) -> torch.Tensor:
+    """
```
snk4tr, line 221: Use raw docstring format 😄 
```suggestion
    r"""
```

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+    Translates image from RGB color space to XYZ
+    Input image in [0, 1] range.
+    """
+
+    mask_below = x <= 0.04045
+    mask_above = x > 0.04045
+
+    tmp = x / 12.92 * mask_below + torch.pow((x + 0.055) / 1.055, 2.4) * mask_above
```
snk4tr, line 229: It seems kinda magical. Worth at least a comment 😄

piq/vsi.py
```diff
@@ -0,0 +1,285 @@
...
+def sdsp(x: torch.Tensor, data_range: float = 255., omega_0: float = 0.021, sigma_f: float = 1.34,
+         sigma_d: float = 145., sigma_c: float = 0.001) -> torch.Tensor:
+    size = x.size()
+
+    size_to_use = (256, 256)
+    x = interpolate(input=x, size=size_to_use, mode='bilinear', align_corners=True)
+    x_lab = rgb2lab(x, data_range=data_range)
+
+    x_fft = torch.fft(torch.stack([x_lab, torch.zeros_like(x_lab)], dim=-1), 2)
+    lg = log_gabor(size_to_use, omega_0, sigma_f).to(x_fft).view(1, 1, *size_to_use, 1)
```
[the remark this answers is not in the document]
denproc: I'll add it to `functional.py`

piq/vsi.py
denproc: The method requires all 3 channels, which are processed differently. One could convert greyscale image to RGB using the best suitable approach for their task.
We can introduce greyscale image processing copying the image three times, but it could create wrong user expectation from the measure. It is a workaround, which is not described in the paper. I would suggest to keep supporting RGB only images.
@zakajd @snk4tr Any other thoughts?

zakajd: I would propose to add support for greyscale images and mention this limitation in docstring.

tests/test_vsi.py
```diff
@@ -0,0 +1,61 @@
...
+        f'Expected finite value for zeros and ones tensos, got {vsi_zeros_ones}'
+
+
+def test_vsi_compare_with_matlab(device) -> None:
+    prediction = torch.tensor(np.array(Image.open('tests/assets/I01.BMP'))).permute(2, 0, 1)
+    target = torch.tensor(np.array(Image.open('tests/assets/i01_01_5.bmp'))).permute(2, 0, 1)
+    predicted_score = vsi(prediction, target, data_range=255, reduction='none')
+    target_score = torch.tensor([0.96405]).to(predicted_score)
+    assert torch.allclose(predicted_score, target_score), f'Expected result similar to MATLAB,' \
+                                                          f'got diff{predicted_score - target_score}'
```
snk4tr, line 43: 👍

PR #134, opened 2020-07-10
Enhancement of SSIM/MS-SSIM and BRISQUE, Refactoring of tests for SSIM/MS-SSIM and BRISQUE

### LINKED ISSUE #118: Randomly failing SSIM test
**Describe the bug**
Test `test_multi_scale_ssim_loss_raise_if_wrong_value_is_estimated_custom_weights` randomly fails.

**To Reproduce**
1. Run tests in you local environment of on GitHub with some [non-related changes](https://github.com/photosynthesis-team/piq/pull/117)
2. Observe the [test failing](https://github.com/photosynthesis-team/piq/runs/819666366) from time to time

**Expected behavior**
Test fails if and only if the implementation of tested functionality is incorrect

**Screenshots**
image

### LINKED ISSUE #126: Big memory consumption during SSIM test
**Describe the bug**
When GPU is available running tests of SSIM will allocate nearly 10GB of GPU memory. 
This leads to test failure if someone is training/inferencing on this GPU at the same time.

**To Reproduce**
1. `pytest tests/test_ssim`
2. Open `nvidia-smi watch -n 1` in a separate window. 
4. See how ~11Gb of GPU memory is allocated during tests

**Expected behavior**
Clear torch cash after running tests, so that maximum memory consumption is limited to tensor size

**Screenshots**
before_test

during_test

denproc:
Closes #126 
Closes #118

## Proposed Changes
- Compare SSIM/MS-SSIM with TensorFlow implementation using real images;
- Compare BRISQUE with another `pybrisque` implementation using real images;
- Refactoring of SSIM/MS-SSIM and BRISQUE;
- Refactoring of tests for SSIM/MS-SSIM and BRISQUE;
- Fixed memory consumption during SSIM/MS-SSIM testing;
- Added better testing on GPU for SSIM/MS-SSIM and BRISQUE.


denproc:
@snk4tr @zakajd ready for review

piq/ssim.py
```diff
@@ -144,66 +142,40 @@ def __init__(self, kernel_size: int = 11, kernel_sigma: float = 1.5, k1: float =
...
-
-        if self.reduction == 'none':
-            return loss
-
-        return {'mean': loss.mean,
-                'sum': loss.sum
-                }[self.reduction](dim=0)
+        score = ssim(x=prediction, y=target, kernel_size=self.kernel_size, kernel_sigma=self.kernel_sigma,
+                     data_range=self.data_range, reduction=self.reduction, full=False, k1=self.k1, k2=self.k2)
+        return torch.ones_like(score) - score
```
snk4tr, line 163: Very neat! Like it 👍

piq/ssim.py
```diff
@@ -433,48 +362,44 @@ def _ssim_per_channel(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, da
...
-    sigma2_sq = compensation * (_gaussian_filter(y * y, kernel) - mu2_sq)
-    sigma12 = compensation * (_gaussian_filter(x * y, kernel) - mu1_mu2)
+    sigma1_sq = compensation * (f.conv2d(x * x, weight=kernel, stride=1, padding=0, groups=n_channels) - mu1_sq)
+    sigma2_sq = compensation * (f.conv2d(y * y, weight=kernel, stride=1, padding=0, groups=n_channels) - mu2_sq)
+    sigma12 = compensation * (f.conv2d(x * y, weight=kernel, stride=1, padding=0, groups=n_channels) - mu1_mu2)
 
     # Set alpha = beta = gamma = 1.
     cs_map = (2 * sigma12 + c2) / (sigma1_sq + sigma2_sq + c2)
-    ssim_map = ((2 * mu1_mu2 + c1) / (mu1_sq + mu2_sq + c1)) * cs_map
+    ssim_map = (2 * mu1_mu2 + c1) / (mu1_sq + mu2_sq + c1) * cs_map
```
snk4tr, line 386: I understand that removing these parentecies are redundant since they do not change the order of operations. However, I would still leave them to make this order more explicit.

piq/ssim.py
```diff
@@ -487,30 +412,44 @@ def _ssim(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, data_range: Un
...
     ssim_val = None
-    for _ in range(levels):
+    for iteration in range(levels):
+        if iteration > 0:
+            padding = (x.shape[2] % 2, x.shape[3] % 2)
+            x = f.pad(x, pad=[padding[0], 0, padding[1], 0], mode='replicate')
+            y = f.pad(y, pad=[padding[0], 0, padding[1], 0], mode='replicate')
+            x = f.avg_pool2d(x, kernel_size=2, padding=0)
+            y = f.avg_pool2d(y, kernel_size=2, padding=0)
+
```
snk4tr, line 449: There is the same code fragment in the `_multi_scale_ssim_complex` function. Consider extracting this duplication into a function (optional)

piq/ssim.py
```diff
@@ -487,30 +412,44 @@ def _ssim(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, data_range: Un
...
     levels = scale_weights_tensor.size(0)
     min_size = (kernel.size(-1) - 1) * 2 ** (levels - 1) + 1
     if x.size(-1) < min_size or x.size(-2) < min_size:
         raise ValueError(f'Invalid size of the input images, expected at least {min_size}x{min_size}.')
 
     mcs = []
     ssim_val = None
-    for _ in range(levels):
+    for iteration in range(levels):
+        if iteration > 0:
```
snk4tr, line 443: Why do you need this condition?

piq/ssim.py
```diff
@@ -625,46 +546,61 @@ def _ssim_complex(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor, data_r
...
-    for _ in range(levels):
-        ssim_val, cs = _ssim_per_channel_complex(x, y, kernel=kernel, data_range=data_range, k1=k1, k2=k2)
-
+    for iteration in range(levels):
         x_real = x[..., 0]
         x_imag = x[..., 1]
         y_real = y[..., 0]
         y_imag = y[..., 1]
-        mcs.append(cs)
+        if iteration > 0:
```
snk4tr, line 580: The same question here: why do you need this condition?

tests/test_ssim.py
```diff
@@ -2,96 +2,81 @@
...
+import numpy as np
+from PIL import Image
 from piq import SSIMLoss, MultiScaleSSIMLoss, ssim, multi_scale_ssim
+from typing import Tuple, List, Any
 
 
 @pytest.fixture(scope='module')
 def prediction() -> torch.Tensor:
-    return torch.rand(3, 3, 256, 256)
+    return torch.rand(3, 3, 161, 161)
```
snk4tr, line 13: Is this a minimal size for MS-SSIM? 
If so, may be add it somewhere in the documentation?

piq/brisque.py
```diff
@@ -11,9 +11,114 @@
 import torch
 from torch.nn.modules.loss import _Loss
 from torch.utils.model_zoo import load_url
-import torch.nn.functional as F
+import torch.nn.functional as f
```
zakajd, line 14: It's very common to import `functional` module as `F`. Why do you want to change this?

tests/test_brisque.py
```diff
@@ -3,96 +3,87 @@
...
 from piq import brisque, BRISQUELoss
+from PIL import Image
+import numpy as np
+from typing import Any
 
 
 @pytest.fixture(scope='module')
 def prediction_grey() -> torch.Tensor:
-    return torch.rand(3, 1, 256, 256)
+    return torch.rand(3, 1, 128, 128)
```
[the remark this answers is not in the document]
zakajd: Yes, I'm using 96x96 now. It works fine for most cases

tests/test_brisque.py
```diff
@@ -3,96 +3,87 @@
...
+    score = brisque(prediction_rgb.to(device), reduction='none', data_range=255.)
+    score_baseline = BRISQUE().get_score(prediction_rgb[0].permute(1, 2, 0).numpy()[..., ::-1])
+    assert torch.isclose(score, torch.tensor(score_baseline).to(score), rtol=1e-3), \
+        f'Expected values to be equal to baseline prediction, got {score.item()} and {score_baseline}'
 
 
-def test_brisque_all_zeros_or_ones() -> None:
-    size = (1, 1, 256, 256)
+def test_brisque_all_zeros(device: Any) -> None:
+    size = (1, 1, 128, 128)
```
zakajd, line 58: Test name is misleading, as you also check tensor of ones inside. 
Consider doing this using `pytest.mark.parametrize`

```suggestion
@pytest.mark.parametrize(
    "input,expectation",
    [(torch.zeros(2, 1, 96, 96), AssertionError),
      (torch.ones(2, 1, 96, 96), AssertionError)],
)
def test_brisque_for_special_cases(input: torch.Tensor, expectation: Any, device: str) -> None:
    with expectation:
        brisque(input.to(device), reduction='mean')

```

piq/utils.py
```diff
@@ -18,7 +18,7 @@ def _adjust_dimensions(input_tensors: Union[torch.Tensor, Tuple[torch.Tensor, to
             tmp = tmp.unsqueeze(0)
         if tmp.dim() != 4 and tmp.dim() != 5:
             raise ValueError(f'Expected 2, 3, 4 or 5 dimensions (got {tensor.dim()})')
-        resized_tensors.append(tmp)
+        resized_tensors.append(tmp.type(torch.float32))
```
zakajd, line 21: Why we need to change tensor type during resize? 
This should be mentioned in docstring

piq/ssim.py
zakajd: To compute SSIM on original scale before downsampling it

tests/test_ssim.py
```diff
@@ -2,96 +2,81 @@
 import itertools
 import pytest
 import tensorflow as tf
-
+import numpy as np
+from PIL import Image
```
zakajd, line 6: Consider using `from skimage.io import imread` instead of `PIL.Image`. It's slightly faster, reads images directly into np.arrays and already included in testing dependencies.

piq/ssim.py
denproc: Just changed order of operation to get rid of final downscale, the logic remains the same.

piq/ssim.py
denproc: The same answer here :)

piq/utils.py
snk4tr: Yeah, I noticed this one as well. 
We have not addressed data types of passed tensors and to me it seems like too much of work to do that now (because then all data types will need to be tested etc.)

I agree with @zakajd that if you change data type it need to be explicitly defined several times in different places or (even better) just be avoided. What is the use case when the absence of this conversion break things down?

tests/test_ssim.py
denproc: Yes, we have limitations due to downscaling inside the MS-SSIM. Added to doc.

piq/utils.py
denproc: This is the case for SSIM and MS-SSIM. We usually use `.to(prediction)`, but it doesn't work here in case of prediction with `dtype=torch.uint8`. For example, it is impossible to create gaussian filter with `dtype=torch.uint8`. 
Next, we will have:
- the rounded result is equal `1` instead of `0.6824` in case of SSIM;
- conflict at `torch.nn.functional.pad` in case of MS-SSIM.

Almost all operations return `float` results. The translation to `torch.float32` is already used as a result of `x = x / float(data_range)`.
Since we implicitly change the type of the input tensors, I propose to make it at `_validation_input` explicitly and show the corresponding warning.

piq/ssim.py
denproc: I suggest to keep it as is. Pytorch 1.6.0 will support complex values, which will lead to possible refactoring of the logic for complex valued methods.

piq/ssim.py
denproc: I suggest to keep it as is. Pytorch 1.6.0 will support complex values, which will lead to possible refactoring of the logic for complex valued methods.

denproc:
### SSIM/MS-SSIM
- I just optimised the code, removed unnecessary functions, iterations and so on. 
- Also, I added padding with reflection mode instead of zero-padding inside `avg_pool2d` and `conv2d`, which makes our implementation closed to TF and original MATLAB implementations.

### Tests
- Tests were optimised to reduce the duplications. It eliminated some inconsistencies.
- Added tests, which use real images.

In general, the implementation is ok. The main sources of possible errors are float precision and zero-paddings instead of reflect/symmetric.

piq/utils.py
snk4tr: Agh I would still avoid doing it.   
If one has `torch.ByteTensor` prediction, let him/her convert it to `torch.float32` and then compute SSIM/MS-SSIM with it. This whole type conversion thing is out of the scope of this lib.

denproc: @snk4tr, `torch.ByteTensor` is `torch.uint8`, which is usual data type for images. Let's take `FSIM`. We can pass `torch.uint8` tensor and got the answer. This is the case for all metrics meaning that we already provide user with such implicit conversion. What is more, it is not logical making one metrics unavailable because of wrong type and keeping other metrics work with it. I move the conversion especially for torch.uint8 inside SSIM.

denproc: @snk4tr @zakajd What do you think about strict assert of the input data type? One can pass the `ByteTensor` to a measure, but they cannot use them for back propagation (gradients are available only for `FloatTensor`).

denproc:
@zakajd @snk4tr ready to merge

snk4tr:
@denproc since #131 was merged you need to resolve conflicts first

denproc:
@snk4tr done

PR #148, opened 2020-07-21
Feature: Mean Deviation Similarity Index (MDSI)

### LINKED ISSUE #113: Metric: Mean Deviation Similarity Index
It was proposed in 2017 an I missed it during initial literature review. It's performance is better that FSIMc or MS-GMSD and it may be good idea to have it in the library.

Copy [Python implementation ](https://github.com/igv/Python-MDSI)
Or copy [original MATLAB implementation](https://www.mathworks.com/matlabcentral/fileexchange/59809)

**Additional context**
Link to original article: https://arxiv.org/pdf/1608.07433.pdf

image

denproc:
Closes #113

## Proposed Changes
- New metric `mdsi`
- New loss function `MDSILoss`
- Tests to validate aforementioned functionality
- Minor changes to other parts

### DIFF
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -1,5 +1,5 @@
 r"""General purpose functions"""
-from typing import Tuple
+from typing import Tuple, Union
 import torch
 
 
@@ -54,3 +54,28 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
     grads = torch.nn.functional.conv2d(x, kernels.to(x), padding=padding)
 
     return torch.sqrt(torch.sum(grads ** 2, dim=-3, keepdim=True))
+
+
+def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
+    r""" Power a 4D tensor with negative values or 5D tensor with complex values.
+    Complex numbers are represented by modulus and argument: r * \exp(i * \phi).
+
+    It will likely to be redundant with introduction of torch.ComplexTensor.
+
+    Args:
+        base: Tensor with shape (B x C x H x W) or (2 x B x C x H x W)
+        exp: int or float
+    Returns:
+        Complex tensor with shape (2 x B x C x H x W)
+    """
+    if base.dim() == 4:
+        x_complex = [base.abs(), torch.atan2(torch.zeros_like(base), base)]
+    elif base.dim() == 5 and base.size(0) == 2:
+        x_complex = [base.pow(2).sum(dim=0).sqrt(), torch.atan2(base[1], base[0])]
+    else:
+        raise ValueError(f'Expected real or complex tensor, got {base.size()}')
+
+    x_complex_pow = [x_complex[0] ** exp, x_complex[1] * exp]
+    x_real_pow = x_complex_pow[0] * torch.cos(x_complex_pow[1])
+    x_imag_pow = x_complex_pow[0] * torch.sin(x_complex_pow[1])
+    return torch.stack((x_real_pow, x_imag_pow), dim=0)
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -57,14 +57,14 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
 
 
 def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
-    r""" Power a 4D tensor with negative values or 5D tensor with complex values.
+    r""" Takes the power of each element in  a 4D tensor with negative values or 5D tensor with complex values.
     Complex numbers are represented by modulus and argument: r * \exp(i * \phi).
 
     It will likely to be redundant with introduction of torch.ComplexTensor.
 
     Args:
         base: Tensor with shape (B x C x H x W) or (2 x B x C x H x W)
-        exp: int or float
+        exp: Exponent
     Returns:
         Complex tensor with shape (2 x B x C x H x W)
     """
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -63,19 +63,22 @@ def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
     It will likely to be redundant with introduction of torch.ComplexTensor.
 
     Args:
-        base: Tensor with shape (B x C x H x W) or (2 x B x C x H x W)
+        base: Tensor with shape (B x C x H x W) or (B x C x H x W x 2)
         exp: Exponent
     Returns:
-        Complex tensor with shape (2 x B x C x H x W)
+        Complex tensor with shape (B x C x H x W x 2)
     """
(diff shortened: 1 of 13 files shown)


denproc:
@zakajd @snk4tr ready for review

piq/mdsi.py
```diff
@@ -0,0 +1,184 @@
...
+    else:
+        raise ValueError(f'Expected combination method "sum" or "mult", got {combination}')
+
+    mct_complex = pow_for_complex(base=gcs, exp=q).mean(dim=(-1, -2), keepdim=True)
+    score = (pow_for_complex(base=gcs, exp=q) - mct_complex).pow(2).sum(dim=0).sqrt()
+    score = (score ** rho).mean(dim=(-1, -2)) ** (o / rho)
+    if reduction == 'none':
+        return score
+    return {'mean': score.mean,
+            'sum': score.sum}[reduction](dim=0)
```
zakajd, line 112: Please check return tensor shape. All other metrics return tensor of size 0, while this metric returns tensor with size 1. 
Minimal example: 
```python
import piq
import torch
prediction = torch.rand(4, 3, 256, 256)
target = torch.rand(4, 3, 256, 256)

piq.mdsi(prediction, target).shape, piq.ssim(prediction, target).shape
```

Result: `(torch.Size([1]), torch.Size([]))`

denproc:
@zakajd done

piq/functional/base.py
```diff
@@ -54,3 +54,28 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
...
+
+
+def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
+    r""" Power a 4D tensor with negative values or 5D tensor with complex values.
+    Complex numbers are represented by modulus and argument: r * \exp(i * \phi).
+
+    It will likely to be redundant with introduction of torch.ComplexTensor.
+
+    Args:
+        base: Tensor with shape (B x C x H x W) or (2 x B x C x H x W)
```
snk4tr, line 66: May you meant `(B x C x H x W x 2)` is case of a complex (5D) tensor?

piq/functional/base.py
```diff
@@ -54,3 +54,28 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
...
+        exp: int or float
```
snk4tr, line 67: Please make a proper description 😄

piq/functional/base.py
```diff
@@ -54,3 +54,28 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
...
+    Returns:
+        Complex tensor with shape (2 x B x C x H x W)
```
snk4tr, line 69: And here about shapes again

piq/functional/colour_conversion.py
```diff
@@ -113,3 +113,23 @@ def rgb2yiq(x: torch.Tensor) -> torch.Tensor:
...
+
+
+def rgb2lhm(x: torch.Tensor) -> torch.Tensor:
+    r"""Convert a batch of RGB images to a batch of LHM images
+
+    Args:
+        x: Batch of 4D (N x 3 x H x W) images in RGB colour space.
+
+    Returns:
+        Batch of 4D (N x 3 x H x W) images in LHM colour space.
```
snk4tr, line 125: Previously you used `B` to denote batch size, but here `N` is used. It would be good to use the same letter for consistency

piq/functional/base.py
denproc: No, the tensors have (2 B C H W) shapes in current implementation.

piq/functional/colour_conversion.py
denproc: However, all functions at `colour conversion` use `N`. I suggest to leave this to separate PR to unify the notation among all files of the library.

zakajd:
> Here is a general suggestion: since we already have several unique Python metrics implementations (previously only MATLAB), it is a great advantage of our library comparing with similar ones. What do you @denproc @zakajd think about adding explicit notes about that somewhere on top of the README?

It's not always true, for example there was already [Python version of MDSI](https://github.com/igv/Python-MDSI). And same with some other metrics, like [BRISQUE](https://github.com/bukalapak/pybrisque),  [VSI](https://github.com/dingkeyan93/IQA-optimization/blob/master/IQA_pytorch/VSI.py), [FSIM](https://github.com/dingkeyan93/IQA-optimization/blob/master/IQA_pytorch/FSIM.py) and VIF [[1](https://github.com/dingkeyan93/IQA-optimization/blob/master/IQA_pytorch/VIF.py) [2](https://github.com/andrewekhalel/sewar/blob/master/sewar/full_ref.py)]. Our advantage is that we support GPU computation, gradients backpropagation etc. 

It seems that really unique are only GMSD and MS-GMSD metrics. In fact something called GMSD do [exist](https://github.com/dingkeyan93/IQA-optimization/blob/master/IQA_pytorch/GMSD.py), but computation is wrong and it should be called Gradient Similarity Measure (GSM)

piq/functional/base.py
denproc: I tried to change the implementation to `(B x C x H x W x 2)`. It leads to reduction in precision and `nan` in gradients.

snk4tr: Then it is a bit weird shape, especially considering that:
1.  For SSIM/MS-SSIM implementations we have complex channels last https://github.com/photosynthesis-team/piq/blob/d3a0540fa2ea80724db11c2b5ecca08a80a7883e/piq/ssim.py#L25
2. It is kind of intuitive to have them last because we go from top to bottom: batches include channels, which consist of height and width pixels, which may be complex i.e. have two values: for real and imaginary parts respectively

@denproc would you consider re-implementing it with complex channels being last?

snk4tr:
@zakajd my main idea was to clearly identify why would anyone bother installing and using our library. All things you listed here (including unique Python implementation of GMSD and MS-GMSD) are perfect examples of such competitive advantages. I will add an issue proposing this since it is important.

zakajd:
@denproc  is it ready to be merged?
What about issue with complex tensors mentioned by Sergey? 

I agree with him that having complex dimension first is not a good idea. In PyTorch complex dimension is supposed to be the last one (for correct work of FFT and other complex operations). As far as I remember we agreed on similar order when you implemented #55

denproc:
@zakajd @snk4tr I looked into it. Making the complex dimension last lead to reduction in precision and `nan` in gradients. Will take some time.

Regarding the description, I would suggest to refactor making description sell the library. It's becoming critical as the library is poorly indexed by search engines. 
One way to improve is creating a more concise list of measures (table with title and link).

denproc:
@zakajd @snk4tr 
The problem with precision was located and led to calculation of the mean values (`mct_complex`). The problem was solved by calculating mean value in two steps independently.
The new `pow_for_complex` calculated with the case precision.
Overall, the precision comparing with MATLAB implementation remains the same.

PR #153, opened 2020-07-24
Optimisation of CI

denproc:
Closes #152 

## Proposed Changes

  - Added `py3.8` testing into CI
  - Removed `py3.7` lint validation

PR #154, opened 2020-07-27
Bugfix: relax the requirements

### LINKED ISSUE #143: [Query] Why is torch 1.5.0 imposed?
In `requirements.txt` the first line `torch>=1.2.0,!=1.5.0` uninstalls existing pytorch version and re-installs this particular version. This could end up in breaking the entire existing pipeline in some case. 
Any particular reason why is this being imposed? 
Is `torch>=1.2.0` already being tried and the results weren't good enough?

denproc:
Closes #143 
The problem was related to the upgrade of the environment with `torch` incompatible with CUDA. 
## Proposed Changes

  - made requirements on `pytorch` and `torchvision` less strict
  - added warning to the BRISQUE score that backprop is not available using `torch==1.5.0'
  - added notes to the docstring and readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,9 @@ prediction = torch.rand(3, 3, 256, 256)
 brisque_index: torch.Tensor = brisque(prediction, data_range=1.)
 ```
 
-In order to use BRISQUE as a loss function, use corresponding PyTorch module:
+In order to use BRISQUE as a loss function, use corresponding PyTorch module. 
+
+Note: the back propagation is not available using `torch==1.5.0`. Update the environment with latest `torch` and `torchvision`.
 ```python
 import torch
 from piq import BRISQUELoss
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -10,14 +10,14 @@ source:
 requirements:
   build:
     - python
-    - pytorch>=1.2.0,!=1.5.0
-    - torchvision>=0.4.0,!=0.6.0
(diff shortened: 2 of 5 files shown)


denproc:
Ready to merge after #153

denproc:
Let me look into it a bit more after @snk4tr comment

denproc:
@snk4tr @zakajd the proposed approach works, ready for review.

As I suggested, requirement `torch>=1.2` covers the tags used to specify the cuda driver version.

Additional context just in case:
```python
[In]: pip list
[Out]: torch                  1.5.0+cu101 
[Out]: torchvision        0.6.0+cu101 

[In]: python setup.py install
[In]: pip list
[Out]: piq                      0.5.1    
[Out]: torch                  1.5.0+cu101 
[Out]: torchvision        0.6.0+cu101 
```

README.md
```diff
@@ -51,7 +51,9 @@ prediction = torch.rand(3, 3, 256, 256)
 brisque_index: torch.Tensor = brisque(prediction, data_range=1.)
 ```
 
-In order to use BRISQUE as a loss function, use corresponding PyTorch module:
+In order to use BRISQUE as a loss function, use corresponding PyTorch module. 
+
+Note: the back propagation is not available using `torch==1.5.0`. Update the environment with latest `torch` and `torchvision`.
```
snk4tr, line 56: Please also add `warnings.warn` at the start of execution of BRISQUE computations if `torch.__version__ == 1.5.0`

PR #162, opened 2020-08-04
bug/gmsd: Add `alpha` parameter from the paper

### LINKED ISSUE #161: Potential bug in MS-GMSDc
**Describe the bug**
It's not a bug report, but more like a potential bug report.
I have a testing pipeline which allows to use metric and compute scores on TID2013 database.

It's common to evaluate metric performance by Spearmank Rank Order Correlation or Kenall Rank Order Correlation.
Values from original paper (mentioned in #67 ):
MS-GMSD - SRCC **0.8139** KROCC **0.6467**
MS-GMSDc - SRCC **0.8895** KROCC **0.7115**
My values: 
MS-GMSD - SRCC **0.811** KROCC **0.643**
MS-GMSDc - SRCC **0.814** KROCC **0.645**

**Expected behavior**
Expected to get values close to original ones. 
Pipeline works fine, because for almost all other metrics I match stated performance

**Additional context**
There is no code from authors to compare with 👎 
So maybe they just did something not mentioned in the paper, thats why results are different

zakajd:
Partially closes #161 

## Proposed Changes

  - After comparing current implementation with original paper I found missing `alpha` term in gradient similarity maps computation. Adding it slightly increases overall performance for both MS-GMSD and MS-GMSDc, but it's still far from values in the paper
  - Paper makes no reference about range of image values used in computations. After changing input range from 1.0 to 255.0 performance dropped by ~30%, so I assume that values should be in [0, 1]
  - There is also no reference about constant value in similarity maps computation. Currently we are using `0.0026` as in original GMSD paper. It's unlikely that it's so important for the final result.

### DIFF
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -30,15 +30,17 @@ def get_meshgrid(size: Tuple[int, int]) -> torch.Tensor:
     return torch.meshgrid(x, y)
 
 
-def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float) -> torch.Tensor:
+def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float, alpha: float = 0.0) -> torch.Tensor:
     r""" Compute similarity_map between two tensors using Dice-like equation.
 
     Args:
         map_x: Tensor with map to be compared
         map_y: Tensor with map to be compared
         constant: Used for numerical stability
+        alpha: Masking coefficient. Substracts - `alpha` * map_x * map_y from denominator and nominator
     """
-    return (2.0 * map_x * map_y + constant) / (map_x ** 2 + map_y ** 2 + constant)
+    return (2.0 * map_x * map_y - alpha * map_x * map_y + constant) / \
+           (map_x ** 2 + map_y ** 2 - alpha * map_x * map_y + constant)
 
 
 def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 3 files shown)


zakajd:
I looked through paper again and couldn't find anything else, that's missing.
@denproc  do you want to take a look at that?

We can merge this small fix, but issue with performance degradation will be left opened.

denproc:
@zakajd 
1. Do you mean GMSD instead of MS_GMSD? In #161 you mentioned GMSD
2. Making he pixel values to [0,255] requires the constant for similarity `T` to be equal 170

zakajd:
@denproc  Ready to be merged after test pass

denproc:
@zakajd, I tested MS-GMSD on our test images. I got the same result for both [0,1] and [0,255] ranges to be used inside the `multi_scale_gmsd`. Also, added `alpha` parameter to `MultiScaleGMSDLoss`. 
Do you get the same results (let's discuss it after your vacation 🙃)?

zakajd:
Hey, @denproc
It took me a bit longer to take a look at this PR. Your changes make results bad again.
MS-GMSD: SRCC 0.812,  KRCC 0.645
MS-GMSDc: SRCC 0.814, KRCC 0.647

This is expected, because there are other range-specific values, such as beta1, beta2 etc.
I changed range to be in [0, 255] for multiscale and it works fine again.
MS-GMSD: SRCC 0.812,  KRCC 0.645
MS-GMSDc: SRCC **0.887**, KRCC **0.710**
Let's keep it this way. Alpha parameter should be the same, because it's range invariant.

@snk4tr ready for review and merge.

snk4tr (approved):

PR #164, opened 2020-08-09
Feature/haar psi

### LINKED ISSUE #122: Metric: Haar Wavelet-Based Perceptual Similarity Index (HaarPSI)
This metric proposed in 2016 inherits best features of VSI and FSIM, top performing full-reference metrics of that time. 
It was also ranked second in recent [comparison on KADID-10k dataset](https://arxiv.org/pdf/1907.02096.pdf)

Original paper: http://www.math.uni-bremen.de/cda/HaarPSI/publications/HaarPSI_preprint_v4.pdf
Python implementation from the authors: https://github.com/rgcda/haarpsi/blob/master/haarPsi.py

**Additional context**
Results from the paper:
image

zakajd:
Closes #122 

## Proposed Changes

  - PyTorch version with less than 1e-4 error when compared to MATLAB


denproc:
@zakajd MATLAB allows to change output style `format long` and display a bit more digits. As a result, we can compare our implementations with better precision.

zakajd:
@denproc Good point. Original implementation used slightly different weights for RGB -> YIQ conversion (with less presision), so accumulated error gives ~1e-5 difference between original implementation and this version.

piq/functional/filters.py
```diff
@@ -1,8 +1,20 @@
...
 import torch
 
 
-def hann_filter(kernel_size) -> torch.Tensor:
+def haar_filter(kernel_size: int) -> torch.Tensor:
+    r"""Creates Haar kernel
+    Returns:
+        kernel: Tensor with shape 1 x `kernel_size` x `kernel_size`"""
+    scale = math.log2(kernel_size)
+    kernel = 2 ** (- scale) * torch.ones((kernel_size, kernel_size))
```
denproc, line 12: No need to import `math`
```suggestion
    kernel =  torch.ones((kernel_size, kernel_size)) / kernel_size
```

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+    haarpsi_zeros_ones = haarpsi(zeros, ones, data_range=1.)
+    assert torch.isfinite(haarpsi_zeros_ones).all(), \
+        f'Expected finite value for zeros and ones tensos, got {haarpsi_zeros_ones}'
+
+
+def test_haarpsi_compare_with_matlab(device: str) -> None:
+    prediction = torch.tensor(np.array(Image.open('tests/assets/I01.BMP'))).permute(2, 0, 1)
+    target = torch.tensor(np.array(Image.open('tests/assets/i01_01_5.bmp'))).permute(2, 0, 1)
+    predicted_score = haarpsi(prediction.to(device), target.to(device), data_range=255, reduction='none')
+    target_score = torch.tensor([0.7170]).to(predicted_score)
```
denproc, line 40: Could you provide the target score with higher precision (even if it does not reduce `atol`)? Another option can be to compare with the original python version.

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+        f'Expected result similar to MATLAB, got diff{predicted_score - target_score}'
+
+
+# ================== Test class: `HaarPSILoss` =================
+def test_haarpsi_loss(input_tensors: Tuple[torch.Tensor, torch.Tensor], device: str) -> None:
+    prediction, target = input_tensors
+    prediction.requires_grad_()
+    loss = HaarPSILoss(data_range=1.)(prediction.to(device), target.to(device))
+    loss.backward()
+    assert prediction.grad is not None, f'Expected finite gradient values after back propagation, got {prediction.grad}'
```
denproc, line 51:
```suggestion
    assert torch.isfinite(prediction.grad).all(), f'Expected finite gradient values after back propagation, got {prediction.grad}'
```

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+
+# ================== Test function: `haarpsi` ==================
+def test_haarpsi_to_be_one_for_identical_inputs(input_tensors: Tuple[torch.Tensor, torch.Tensor], device: str) -> None:
+    prediction, _ = input_tensors
+    index = haarpsi(prediction.to(device), prediction.to(device), data_range=1., reduction='none')
+    index_255 = haarpsi(prediction.to(device) * 255, prediction.to(device) * 255, data_range=255, reduction='none')
+    assert torch.isclose(index, torch.ones_like(index, device=device)).all(), \
+        f'Expected index to be equal 1, got {index}'
+    assert torch.isclose(index_255, torch.ones_like(index_255, device=device)).all(), \
+        f'Expected index to be equal 1, got {index_255}'
```
denproc, line 16:
```suggestion
    assert torch.allclose(index, torch.ones_like(index, device=device)), \
        f'Expected index to be equal 1, got {index}'
    assert torch.allclose(index_255, torch.ones_like(index_255, device=device)), \
        f'Expected index to be equal 1, got {index_255}'
```

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_haarpsi_symmetry(input_tensors: Tuple[torch.Tensor, torch.Tensor], device: str) -> None:
+    prediction, target = input_tensors
+    result = haarpsi(prediction.to(device), target.to(device), data_range=1., reduction='none')
+    result_sym = haarpsi(target.to(device), prediction.to(device), data_range=1., reduction='none')
+    assert torch.isclose(result_sym, result).all(), f'Expected the same results, got {result} and {result_sym}'
```
denproc, line 23:
```suggestion
    assert torch.allclose(result_sym, result), f'Expected the same results, got {result} and {result_sym}'
```

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+    loss.backward()
+    assert prediction.grad is not None, f'Expected finite gradient values after back propagation, got {prediction.grad}'
+
+
+def test_haarpsi_loss_zero_for_equal_input(input_tensors: Tuple[torch.Tensor, torch.Tensor], device: str) -> None:
+    prediction, _ = input_tensors
+    target = prediction.clone()
+    prediction.requires_grad_()
+    loss = HaarPSILoss(data_range=1.)(prediction.to(device), target.to(device))
+    assert torch.isclose(loss, torch.zeros_like(loss), atol=1e-5), \
```
denproc, line 59: Why do we have different precision compared to `test_haarpsi_to_be_one_for_identical_inputs`?

tests/test_haarpsi.py
```diff
@@ -0,0 +1,60 @@
...
+    haarpsi_ones = haarpsi(ones, ones, data_range=1.)
+    assert torch.isfinite(haarpsi_ones).all(), f'Expected finite value for ones tensos, got {haarpsi_ones}'
+    haarpsi_zeros_ones = haarpsi(zeros, ones, data_range=1.)
+    assert torch.isfinite(haarpsi_zeros_ones).all(), \
+        f'Expected finite value for zeros and ones tensos, got {haarpsi_zeros_ones}'
+
+
+def test_haarpsi_compare_with_matlab(device: str) -> None:
+    prediction = torch.tensor(np.array(Image.open('tests/assets/I01.BMP'))).permute(2, 0, 1)
+    target = torch.tensor(np.array(Image.open('tests/assets/i01_01_5.bmp'))).permute(2, 0, 1)
```
denproc, line 38: Let's stick with `skimage` instead of `PIL.Image`.

piq/haarpsi.py
```diff
@@ -0,0 +1,204 @@
...
+        [1] R. Reisenhofer, S. Bosse, G. Kutyniok & T. Wiegand (2017)
+            'A Haar Wavelet-Based Perceptual Similarity Index for Image Quality Assessment'
+            http://www.math.uni-bremen.de/cda/HaarPSI/publications/HaarPSI_preprint_v4.pdf
+        [2] Code from authors on MATLAB and Python
+            https://github.com/rgcda/haarpsi
+    """
+
+    _validate_input(input_tensors=(x, y), allow_5d=False, scale_weights=None)
+    x, y = _adjust_dimensions(input_tensors=(x, y))
+
```
denproc, line 55: Add assertion to validate the size of the image to be greater than max possible `kernel size = 2 ** (scale + 1)`

piq/haarpsi.py
```diff
@@ -0,0 +1,204 @@
...
+        upper_pad = kernel_size // 2 - 1
+        bottom_pad = kernel_size // 2
+        pad_to_use = [upper_pad, bottom_pad, upper_pad, bottom_pad]
+        coeff_x = torch.nn.functional.conv2d(F.pad(x_lum, pad=pad_to_use, mode='constant'), kernels.to(x))
+        coeff_y = torch.nn.functional.conv2d(F.pad(y_lum, pad=pad_to_use, mode='constant'), kernels.to(y))
+    
+        coefficients_x.append(coeff_x)
+        coefficients_y.append(coeff_y)
+
+    # Shape B x 6 x H x W
```
denproc, line 101: Add a comment describing the second dimension.

tests/test_haarpsi.py
zakajd: Because current implementation of `logit` function is not numerically stable at the borders, and near zero it's has bigger error, than near ones.
Starting from PyTorch 1.7.0 there will be built-in function `torch.logit` which handles values close to 0 and 1 correctly.

tests/test_haarpsi.py
zakajd: Done 👍

zakajd:
@denproc  Ready for review
We have some problems with Python 3.8 testing, it fails during packages collection. Other versions work fine.

tests/test_haarpsi.py
denproc: Current implementation performs the `logit` near `1.` and then is substituted from 1. Therefore there should be the same result, shouldn't it?

zakajd: I noticed, that when passing two tensors with zeros error is bigger, than when passing two tensors with ones. Not sure what exactly causes such behavior.
Original MATLAB and Python versions just fail for such inputs, so having `1e-4` error is not that bad.

denproc: The difference is due to `rtol=1e05` of the `torch.isclose`.  Resolved.

piq/haarpsi.py
```diff
@@ -0,0 +1,210 @@
...
+        y = F.pad(y, pad=pad_to_use)
+
+        x = F.avg_pool2d(x, kernel_size=2, stride=2, padding=0)
+        y = F.avg_pool2d(y, kernel_size=2, stride=2, padding=0)
+
+    num_channels = x.size(1)
+    # Convert RGB to YIQ color space https://en.wikipedia.org/wiki/YIQ
+    if num_channels == 3:
+        x = rgb2yiq(x)
+        y = rgb2yiq(y)
```
denproc, line 81: The images should be translated to YIQ and then subsampled according to original implementation.

tests/test_haarpsi.py
```diff
@@ -0,0 +1,68 @@
...
+        haarpsi(prediction, target, data_range=1.)
+
+
+def test_haarpsi_compare_with_matlab(device: str) -> None:
+    prediction = torch.tensor(imread('tests/assets/I01.BMP')).permute(2, 0, 1)
+    target = torch.tensor(imread('tests/assets/i01_01_5.bmp')).permute(2, 0, 1)
+    predicted_score = haarpsi(prediction.to(device), target.to(device), data_range=255, reduction='none')
+    target_score = torch.tensor([0.71706527]).to(predicted_score)
+    assert torch.isclose(predicted_score, target_score, atol=1e-4),\
+        f'Expected result similar to MATLAB, got diff{predicted_score - target_score}'
```
denproc, line 49: I tried to use less precise weights to convert to YIQ (same as in original) resulting in worse performance. Where do we lose the precision then (despite the order of subsampling and YIQ conversion)?

denproc: Looks like the result matches the MATLAB with their weights of YIQ and right order of subsampling.

zakajd:
@denproc I changed the order, but error is left the same.

Issue #133, opened 2020-07-10
Add support for negative tensors


zakajd:
**Is your feature request related to a problem? Please describe.**
Currently we assert that input tensors are positive in `_validate_input` function. This is reasonable, but for several metrics in some cases input CAN be negative, for example if it's already normalized imaged.

**Describe the solution you'd like**
Find a clever way to adjust current validation

zakajd:
Closing because #169 solved the issue.

PR #169, opened 2020-09-30
Introduce `allow_negative` flag in tensor check

### LINKED ISSUE #168: Reasons for assert torch.all(tensor >= 0)
Inside the _validate_input function, there is a check for values of the input tensor to be >=0: https://github.com/photosynthesis-team/piq/blob/master/piq/utils/common.py#L46

However, this introduces difficulties when using some of the losses from piq. For example, I would like to use DISTS as a loss function for training a neural network without an activation function at the end. 

Could you please clarify the reasons why this check is required?

zakajd:
Closes #168 and relevant to #133 

## Proposed Changes
  - Flag `allow_negative`
  - Minor documentation update

### DIFF
--- a/piq/perceptual.py
+++ b/piq/perceptual.py
@@ -75,10 +75,11 @@ EPS = 1e-10
 
 
 class ContentLoss(_Loss):
-    r"""Creates Content loss that can be used for image style transfer of as a measure in
-    image to image tasks.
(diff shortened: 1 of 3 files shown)


snk4tr:
@zakajd conflicts need to be resolved.

zakajd:
@snk4tr  ready to be merged

snk4tr (approved):

PR #171, opened 2020-09-30
Fix bug in GMSD test

### LINKED ISSUE #170: GMSD test fails
**Describe the bug**
One of the tests fail due to an error in it's logic
```
def test_gmsd_supports_different_data_ranges(prediction: torch.Tensor, target: torch.Tensor, device: str) -> None:
    prediction_255 = (prediction * 255).type(torch.uint8)
    target_255 = (target * 255).type(torch.uint8)
    measure = gmsd(prediction.to(device), target.to(device))

    measure_255 = gmsd(prediction_255.to(device), target_255.to(device), data_range=255)
    diff = torch.abs(measure_255 - measure)
    assert diff <= 1e-4, f'Result for same tensor with different data_range should be the same, got {diff}'
```

Casting to `torch.uint8` introduces rounding error, which results in different scores.

**Expected behavior**
Correct comparison on different ranges.

zakajd:
Closes #170


denproc (approved):
Nice catch

PR #180, opened 2020-10-30
Add mypy test to CI

### LINKED ISSUE #160: Add `mypy` test to CI
We are currently writing type hints for all functions and classes, but never actually use them in testing pipeline.
It can be done by running `python3 -m mypy .` and now this produces dozens of errors. Some of them are valid bugs and using this information can be helpfull

- Fix current errors
- Add `mypy` to CI pipeline

Leave everything as is and use type hints only as a reference 

**Additional context**
image

zakajd:
Closes #160 

## Proposed Changes

  - `mypy.ini` with proper config
  - Fix existing errors
  - Add test to CI 

I had to add "--allow-redefinition" flag to ignore errors due to re-use of the same variable in the functions multiple times.
PyTorch is currently in progress of adding proper types, but currently there a lot of errors, so I ignore them (see `mypy.ini` file for details).

### DIFF
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -7,7 +7,7 @@ Credits:
     https://live.ece.utexas.edu/research/Quality/index_algorithms.htm BRISQUE
     https://github.com/bukalapak/pybrisque
 """
-from typing import Union, Tuple
+from typing import Union, Tuple, List
 import warnings
 import torch
 from torch.nn.modules.loss import _Loss
@@ -191,7 +191,7 @@ def _natural_scene_statistics(luma: torch.Tensor, kernel_size: int = 7, sigma: f
     std = ((std - mu_sq).abs().sqrt())
 
     luma_nrmlzd = (luma - mu) / (std + C)
-    features = []
+    features: List = []
     alpha, sigma = _ggd_parameters(luma_nrmlzd)
     features.extend((alpha, sigma.pow(2)))
 
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -7,7 +7,7 @@ Credits:
     https://live.ece.utexas.edu/research/Quality/index_algorithms.htm BRISQUE
     https://github.com/bukalapak/pybrisque
 """
-from typing import Union, Tuple, List
+from typing import Union, Tuple
 import warnings
 import torch
 from torch.nn.modules.loss import _Loss
@@ -191,9 +191,8 @@ def _natural_scene_statistics(luma: torch.Tensor, kernel_size: int = 7, sigma: f
     std = ((std - mu_sq).abs().sqrt())
 
     luma_nrmlzd = (luma - mu) / (std + C)
-    features: List = []
     alpha, sigma = _ggd_parameters(luma_nrmlzd)
-    features.extend((alpha, sigma.pow(2)))
+    features = [alpha, sigma.pow(2)]
 
     shifts = [(0, 1), (1, 0), (1, 1), (-1, 1)]
 
--- a/piq/fsim.py
+++ b/piq/fsim.py
@@ -6,7 +6,7 @@ References:
 """
 import math
 import functools
-from typing import Union, Tuple
+from typing import Union, Tuple, List
 
 import torch
 from torch.nn.modules.loss import _Loss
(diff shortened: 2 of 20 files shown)


zakajd:
@snk4tr Ready for review

piq/brisque.py
```diff
@@ -191,7 +191,7 @@ def _natural_scene_statistics(luma: torch.Tensor, kernel_size: int = 7, sigma: f
     std = ((std - mu_sq).abs().sqrt())
 
     luma_nrmlzd = (luma - mu) / (std + C)
-    features = []
+    features: List = []
```
snk4tr, line 194: I think this one is a bit redundant since it can be nothing then a list. 
Would you consider removing this type annotation?

piq/fsim.py
```diff
@@ -177,7 +177,7 @@ def _construct_filters(x: torch.Tensor, scales: int = 4, orientations: int = 4,
     lp = _lowpassfilter(size=(H, W), cutoff=.45, n=15)
 
     # Construct the radial filter components...
-    log_gabor = []
+    log_gabor: List = []
```
snk4tr, line 180: Same here

piq/fsim.py
```diff
@@ -187,7 +187,7 @@ def _construct_filters(x: torch.Tensor, scales: int = 4, orientations: int = 4,
         log_gabor.append(gabor_filter)
 
     # Then construct the angular filter components...
-    spread = []
+    spread: List = []
```
snk4tr, line 190: Same here

piq/functional/colour_conversion.py
```diff
@@ -54,7 +54,7 @@ def xyz2lab(x: torch.Tensor, illuminant: str = 'D50', observer: str = '2') -> to
     """
     epsilon = 0.008856
     kappa = 903.3
-    illuminants = \
+    illuminants: Dict = \
```
snk4tr, line 57: Same here

piq/functional/layers.py
```diff
@@ -16,7 +17,7 @@ def __init__(self, kernel_size: int = 3, stride: int = 2, padding=1) -> None:
         self.stride = stride
         self.padding = padding
 
-        self.kernel = None
+        self.kernel: Optional[torch.Tensor] = None
```
snk4tr, line 20: Here I think it is good to have type annotation here because it shows target data type

piq/brisque.py
zakajd: Yes. 
I've experimented with inline typing and forgot to remove this. Anything else?

piq/kid.py
```diff
@@ -79,8 +79,8 @@ def _mmd2_and_variance(
         diag_X = torch.diagonal(K_XX)
         diag_Y = torch.diagonal(K_YY)
 
-        sum_diag_X = diag_X.sum()
-        sum_diag_Y = diag_Y.sum()
+        sum_diag_X = torch.sum(diag_X)
+        sum_diag_Y = torch.sum(diag_Y)
```
snk4tr, line 83: I am ok with this change but what is the difference?

piq/perceptual.py
```diff
@@ -104,7 +104,7 @@ class ContentLoss(_Loss):
         https://arxiv.org/abs/1801.03924
     """
 
-    def __init__(self, feature_extractor: Union[str, Callable] = "vgg16", layers: Tuple[str] = ("relu3_3", ),
+    def __init__(self, feature_extractor: Union[str, torch.nn.Module] = "vgg16", layers: Iterable[str] = ("relu3_3", ),
```
snk4tr, line 107: I would leave `Callable` instead of `torch.nn.Module` since it is possible that a user would want to use some function which not only includes forward through a `torch.nn.Module` extractor but also some pre- or post-processing

piq/kid.py
zakajd: `torch.sum()` has correct typing info. `sum()` doesn't and throws an error

piq/functional/colour_conversion.py
zakajd: I'll update this to `Dict[str, Dict]` because otherwise `mypy` can't determine correct type for dict of dicts.

piq/perceptual.py
zakajd: Just a function won't work in current implementation, because we have calls like:
`self.model.parameters()`, `self.model.to(prediction)`, `self.model._modules`. Easiest way to add custom preprocessing is inherit class and modify `forward` step.

piq/kid.py
snk4tr: Nice to know. Thanks 😄

piq/perceptual.py
snk4tr: Ok, makes sense. Let's then leave it this way.

snk4tr (approved):

PR #184, opened 2020-11-09
Metric: PieAPP

### LINKED ISSUE #181: Add metric PieAPP
Proposed in [PieAPP: Perceptual Image-Error Assessment through Pairwise Preference](https://arxiv.org/pdf/1806.02067.pdf) paper this method showed great performance on [PIPAL dataset](https://arxiv.org/pdf/2007.12142.pdf) which is the largest and most diverse dataset today (as far as I know). 

Re-implement model using original code, simplify and unify calling interface with DISTS and LPIPS metrics.

**Additional context**
[Code from authors](https://github.com/prashnani/PerceptualImageError)
[Link to model weights](https://web.ece.ucsb.edu/~ekta/projects/PieAPPv0.1/weights/PieAPPv0.1.pth)

zakajd:
Closes #181 

Original code is badly written, so I changed it in a few places. 
Model structure is similar to perceptual losses, but small differences required to overwrite every call, so I decided not to inherit from `ContentLoss`.

## Proposed Changes
- [x] Metric re-implementation with equal performance to original version
- [x] Updated docs
- [x] Add model weights to last release 
- [x] Tests

### DIFF
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -87,7 +87,7 @@ def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
 
 
 def crop_patches(x: torch.Tensor, size=64, stride=32) -> torch.Tensor:
-    r"""Crop tensor with images into small patches 
+    r"""Crop tensor with images into small patches
     Args:
         images: Tensor with shape (N, C, H, W)
         size: Size of a square patch
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -93,6 +93,8 @@ def crop_patches(x: torch.Tensor, size=64, stride=32) -> torch.Tensor:
         size: Size of a square patch
         stride: Step between patches
     """
+    assert (x.shape[2] >= size) and (x.shape[3] >= size), \
+        f"Images must be bigger than patch size. Got ({x.shape[2], x.shape[3]}) and ({size}, {size})"
     channels = x.shape[1]
     patches = x.unfold(1, channels, channels).unfold(2, size, stride).unfold(3, size, stride)
     patches = patches.reshape(-1, channels, size, size)
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -59,7 +59,7 @@ def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
 
 
 def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
-    r""" Takes the power of each element in  a 4D tensor with negative values or 5D tensor with complex values.
+    r""" Takes the power of each element in a 4D tensor with negative values or 5D tensor with complex values.
     Complex numbers are represented by modulus and argument: r * \exp(i * \phi).
 
     It will likely to be redundant with introduction of torch.ComplexTensor.
@@ -84,3 +84,16 @@ def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
     x_real_pow = x_complex_pow_r * torch.cos(x_complex_pow_phi)
     x_imag_pow = x_complex_pow_r * torch.sin(x_complex_pow_phi)
     return torch.stack((x_real_pow, x_imag_pow), dim=-1)
+
+
+def crop_patches(x: torch.Tensor, size=64, stride=32) -> torch.Tensor:
+    r"""Crop tensor with images into small patches 
+    Args:
+        images: Tensor with shape (N, C, H, W)
+        size: Size of a square patch
+        stride: Step between patches
+    """
+    channels = x.shape[1]
+    patches = x.unfold(1, channels, channels).unfold(2, size, stride).unfold(3, size, stride)
+    patches = patches.reshape(-1, channels, size, size)
+    return patches
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -89,7 +89,7 @@ def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
 def crop_patches(x: torch.Tensor, size=64, stride=32) -> torch.Tensor:
     r"""Crop tensor with images into small patches
     Args:
-        images: Tensor with shape (N, C, H, W)
+        x: Tensor with shape (N, C, H, W), expected to be images-like entities
         size: Size of a square patch
         stride: Step between patches
     """
--- a/piq/pieapp.py
+++ b/piq/pieapp.py
@@ -18,27 +18,27 @@ from piq.functional import crop_patches
 
 class PieAPPModel(nn.Module):
     """Model used for PieAPP score computation
-    Args:
-        num_features: Base feature size, which is multiplied by 2 every 2 blocks
     """
-    def __init__(self, features=64):
+    # Base feature size, which is multiplied by 2 every 2 blocks
+    FEATURES = 64
+    
+    def __init__(self):
         super().__init__()
 
-        self.features = features
         self.pool = nn.MaxPool2d(2, 2)
         self.flatten = nn.Flatten(start_dim=1)
 
-        self.conv1 = nn.Conv2d(3, features, kernel_size=3, padding=1)
-        self.conv2 = nn.Conv2d(features, features, kernel_size=3, padding=1)
-        self.conv3 = nn.Conv2d(features, features, kernel_size=3, padding=1)
-        self.conv4 = nn.Conv2d(features, features * 2, kernel_size=3, padding=1)
-        self.conv5 = nn.Conv2d(features * 2, features * 2, kernel_size=3, padding=1)
-        self.conv6 = nn.Conv2d(features * 2, features * 2, kernel_size=3, padding=1)
-        self.conv7 = nn.Conv2d(features * 2, features * 4, kernel_size=3, padding=1)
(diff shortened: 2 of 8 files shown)


snk4tr:
@zakajd thanks a lot for your contribution! I will have a look tomorrow morning.

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        self.conv4 = nn.Conv2d(features, features * 2, kernel_size=3, padding=1)
+        self.conv5 = nn.Conv2d(features * 2, features * 2, kernel_size=3, padding=1)
+        self.conv6 = nn.Conv2d(features * 2, features * 2, kernel_size=3, padding=1)
+        self.conv7 = nn.Conv2d(features * 2, features * 4, kernel_size=3, padding=1)
+        self.conv8 = nn.Conv2d(features * 4, features * 4, kernel_size=3, padding=1)
+        self.conv9 = nn.Conv2d(features * 4, features * 4, kernel_size=3, padding=1)
+        self.conv10 = nn.Conv2d(features * 4, features * 8, kernel_size=3, padding=1)
+        self.conv11 = nn.Conv2d(features * 8, features * 8, kernel_size=3, padding=1)
+
+        self.fc1_score = nn.Linear(in_features=120832, out_features=512, bias=True)
```
snk4tr, line 43: I guess this line makes having `features` as a parameter senseless because number of input features is hardcoded. I also think that since you use pretrained model weights you can't really change parameters.

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        # Term for numerical stability
+        self.EPS = 1e-6
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        r"""
+        Forward pass batch of square patches with shape  (N, C, features, features)
+        Returns:
+            
+        """
+        assert x.shape[2] == x.shape[3] == self.features, \
```
snk4tr, line 58: Since this model is public and can be applied by a user for feature extraction, I also propose to check that the passed tensor meets other requirements (non-negative - if required? 4D etc.)

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+            f"Expected square input with shape {self.features, self.features}"
```
snk4tr, line 59:
```suggestion
            f"Expected square input with shape {self.features, self.features}, got {x.shape}"
```

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        self.ref_score_subtract = nn.Linear(in_features=1, out_features=1, bias=True)
+
+        # Term for numerical stability
+        self.EPS = 1e-6
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        r"""
+        Forward pass batch of square patches with shape  (N, C, features, features)
+        Returns:
+
```
snk4tr, line 56: I think this part of the doctoring is missed

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        self.fc1_weight = nn.Linear(in_features=2048, out_features=512)
+        self.fc2_weight = nn.Linear(in_features=512, out_features=1, bias=True)
+        self.ref_score_subtract = nn.Linear(in_features=1, out_features=1, bias=True)
+
+        # Term for numerical stability
+        self.EPS = 1e-6
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        r"""
+        Forward pass batch of square patches with shape  (N, C, features, features)
```
snk4tr, line 54: Not sure but seems like it should be this way:
```suggestion
        Forward pass a batch of square patches with shape  (N, C, self.features, self.features)
```

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        x11 = self.flatten(F.relu(self.conv11(self.pool(F.relu(self.conv10(x9))))))
+        # flatten and concatenate
+        features = torch.cat((self.flatten(x3), self.flatten(x5), self.flatten(x7), self.flatten(x9), x11), dim=1)
+        return features, x11
+
+    def compute_difference(self, features_diff, weights_diff):
+        r"""
+        Args:
+            features_diff: Tensor of shape (N * NUM_PATCHES, C_1)
+            features_diff: Tensor of shape (N * NUM_PATCHES, C_2)
```
snk4tr, line 79: What is `NUM_PATCHES`? I don't see this global variable in the module

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        self.model.to(device=prediction.device)
+        prediction_features, prediction_weights = self.get_features(prediction)
+        target_features, target_weights = self.get_features(target)
+
+        distances, weights = self.model.compute_difference(
+            target_features - prediction_features,
+            target_weights - prediction_weights
+        )
+
+        # Shape (N, NUM_PATCHES)
```
snk4tr, line 167: What is `NUM_PATCHES`? I don't see this variable in the module

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+class PieAPP(_Loss):
+    r"""
+    Implementation of Perceptual Image-Error Assessment through Pairwise Preference.
+    
+    Expects input to be in range [0, 1] with no normalization and RGB channel order.
+    Input images are croped into smaller patches and final score is mean of patch scores.
+    
+
+    Args:
+        reduction: Reduction over samples in batch: "mean"|"sum"|"none".
```
snk4tr, line 101: If `final score is mean of patch scores`, then I think `reduction` parameters is a bit redundant. We either should avoid explicit description in the doctoring (which would be weird if this is a part of an algorithm for the metric) or remove other redundancy options. What do you think?

piq/pieapp.py
```diff
@@ -0,0 +1,205 @@
...
+        
+        Returns:
+            features: List of features extracted from intermediate layers
+        """
+        # Rescale to [0, 255] range on which models was trained
+        x = (x / float(self.data_range)) * 255
+
+        x_patches = crop_patches(x, size=64, stride=self.stride)
+
+        features, weights = self.model(x_patches)
```
snk4tr, line 194: What do you think about adding this line? May be useful when PieAPP is used as metric.
```suggestion
        with torch.no_grad():
            features, weights = self.model(x_patches)
```

piq/functional/base.py
```diff
@@ -84,3 +84,16 @@ def pow_for_complex(base: torch.Tensor, exp: Union[int, float]) -> torch.Tensor:
     x_real_pow = x_complex_pow_r * torch.cos(x_complex_pow_phi)
     x_imag_pow = x_complex_pow_r * torch.sin(x_complex_pow_phi)
     return torch.stack((x_real_pow, x_imag_pow), dim=-1)
+
+
+def crop_patches(x: torch.Tensor, size=64, stride=32) -> torch.Tensor:
```
snk4tr, line 89: What happens if size < x[-1] (or x[-2])?  
I think we need a test to cover this case.

snk4tr (changes requested):
According to the codecov, `replace_pooling` functionality is not covered with tests. Could you add a test for the `PieAPPModel` that would verify that layers are replaces and model still works after this change?

Overall, this is a great PR! Only minor changes are required.

piq/pieapp.py
zakajd: Added `torch.autograd.set_grad_enabled` context manager and additional flag in `init`. 
Some users may want to use PieAPP as a loss, so turning off all gradients isn't good idea.

piq/pieapp.py
zakajd: Updated docstring description. Final score for _individual_ image  is mean of it's patch scores. So we still may want to reduce results over batch.

piq/pieapp.py
zakajd: Removed this comment. 
Number of patches depend on input image size, crop size and stride, it's not some fixed value.

piq/pieapp.py
zakajd: Agree. Removed it from __init__ and hardcoded for now.
As soon as dataset used for PieAPP model training will be released we can reconsider this implementation

piq/pieapp.py
```diff
@@ -0,0 +1,192 @@
...
+        enable_grad: Flag to compute gradients. Usefull when PieAPP used as a loss. Default: False.
+    
+    References:
+        .. [1] Ekta Prashnani, Hong Cai, Yasamin Mostofi, Pradeep Sen
+            (2018). PieAPP: Perceptual Image-Error Assessment through Pairwise Preference
+            https://arxiv.org/abs/1806.02067
+        .. [2] https://github.com/prashnani/PerceptualImageError
+
+    """
+    # TODO(zakajd) 10/12/2020: Load weights to release and change this link
```
zakajd, line 117: @snk4tr this PR is ready for merge if you don't have additional comments.

I'm currently in the place with very slow internet, so it would be great if you also upload model weights to last release and change the link in code.

piq/pieapp.py
snk4tr: I also added a TODO comment to not forget

piq/pieapp.py
snk4tr: Done

snk4tr (approved):
Nicely done, @zakajd!

PR #186, opened 2020-11-24
Creating eye matrix right on target device, avoids excessive cpu-load on many-core systems

akamaus:
Currently FID calculations on GPU still result in excessive CPU load (around 3000% on 72 core system). Immediate culprit is torch.eye which uses weird parallelization strategy. See https://github.com/pytorch/pytorch/issues/48251 

## Proposed Changes
- It's better to create eye matrix in place, this saves data transfer latency (always) and avoids excessive cpu load (untill torch bug is fixed)


zakajd:
Hi, thank you for PR!
Didn't see such behaviour in older versions of PyTorch, probably something broke after 1.6 release. Will investigate and merge  your solution in a few days.

zakajd (approved):
Approve 🆗

PR #191, opened 2020-12-08
Update fid_inception.py

bes-dev:
fix for case when we use normalize_mode=False


zakajd (changes requested):
Thank you for PR 👍 

Can you also add test to `tests/test_fid` that will show that FID works with both `normalize=True` and `normalize=False`?

bes-dev:
@zakajd sure. Is it looks good for you now?

zakajd (approved):

PR #195, opened 2020-12-10
Add test for correct data range

### LINKED ISSUE #194: Incorrect support for different data ranges
**Describe the bug**
Currently in different full reference metrics we have different scaling procedures, which results in some hidden bugs.
For example, `vif_p` with `data_range=128` will fail. Also we never check, that values in the given tensors are actually lower than the given data range

**Expected behavior**
Add 2 test to all metrics:
1. Check that result for same tensor scaled to different data ranges is the same
2. Check that if values in the tensor are bigger, than data range it will cause AssertionError

zakajd:
Closes #194 and #196 

## Proposed Changes

  - Test that values are always lower than `data_range`
  - Test that results is consistent for different data ranges

## Progres
- [x] BRISQUE
- [x] FSIM
- [x] GMSD
- [x] HaarPSI
- [x] MDSI
- [x] PSNR
- [x] SSIM
- [x] VIF
- [x] VSI 
- [x] PieAPP

### DIFF
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -51,7 +51,7 @@ def brisque(x: torch.Tensor,
     _validate_input(input_tensors=x, allow_5d=False, kernel_size=kernel_size, data_range=data_range)
     x = _adjust_dimensions(input_tensors=x)
 
-    x = x / float(data_range) * 255
+    x = x / data_range * 255
 
     if x.size(1) == 3:
         x = rgb2yiq(x)[:, :1]
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -48,11 +48,10 @@ def brisque(x: torch.Tensor,
                       f'More info is available at https://github.com/photosynthesis-team/piq/pull/79 and'
                       f'https://github.com/pytorch/pytorch/issues/38869.')
 
-    _validate_input(input_tensors=x, allow_5d=False, kernel_size=kernel_size)
+    _validate_input(input_tensors=x, allow_5d=False, kernel_size=kernel_size, data_range=data_range)
     x = _adjust_dimensions(input_tensors=x)
 
-    assert data_range >= x.max(), f'Expected data range greater or equal maximum value, got {data_range} and {x.max()}.'
-    x = x * 255. / data_range
+    x = x / float(data_range) * 255
 
     if x.size(1) == 3:
         x = rgb2yiq(x)[:, :1]
--- a/piq/gmsd.py
+++ b/piq/gmsd.py
@@ -47,7 +47,7 @@ def gmsd(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean'
         input_tensors=(prediction, target), allow_5d=False, scale_weights=None, data_range=data_range)
     prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
 
-    # Rescale to [0, 1] range    
+    # Rescale to [0, 1] range
     prediction = prediction / float(data_range)
     target = target / float(data_range)
 
--- a/piq/gmsd.py
+++ b/piq/gmsd.py
@@ -43,9 +43,11 @@ def gmsd(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean'
         https://arxiv.org/pdf/1308.3052.pdf
     """
 
-    _validate_input(input_tensors=(prediction, target), allow_5d=False, scale_weights=None)
+    _validate_input(
+        input_tensors=(prediction, target), allow_5d=False, scale_weights=None, data_range=data_range)
     prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
 
+    # Rescale to [0, 1] range    
     prediction = prediction / float(data_range)
     target = target / float(data_range)
 
@@ -179,7 +181,8 @@ def multi_scale_gmsd(prediction: torch.Tensor, target: torch.Tensor, data_range:
     Returns:
         Value of MS-GMSD. 0 <= GMSD loss <= 1.
     """
-    _validate_input(input_tensors=(prediction, target), allow_5d=False, scale_weights=scale_weights)
+    _validate_input(
+        input_tensors=(prediction, target), allow_5d=False, scale_weights=scale_weights, data_range=data_range)
     prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
 
     # Values from the paper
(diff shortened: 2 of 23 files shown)


zakajd:
Let's merge this first, then #184

snk4tr:
> Let's merge this first, then #184

Well, I think I missed this one. #184 is already merged. Is it a problem? @zakajd do you need to add something now (probably one more test from the PieAPP metric) to make this PR complete?

zakajd:
Yes, I need to add data range tests to PieAPP before merge. 
Will do it tomorrow and merge if you don't have any additional questions

snk4tr:
> Yes, I need to add data range tests to PieAPP before merge.
> Will do it tomorrow and merge if you don't have any additional questions

I will add my review now and revise tomorrow evening after the data range tests for the PieAPP will be added. Then we can merge 😄

piq/brisque.py
```diff
@@ -48,11 +48,10 @@ def brisque(x: torch.Tensor,
                       f'More info is available at https://github.com/photosynthesis-team/piq/pull/79 and'
                       f'https://github.com/pytorch/pytorch/issues/38869.')
 
-    _validate_input(input_tensors=x, allow_5d=False, kernel_size=kernel_size)
+    _validate_input(input_tensors=x, allow_5d=False, kernel_size=kernel_size, data_range=data_range)
     x = _adjust_dimensions(input_tensors=x)
 
-    assert data_range >= x.max(), f'Expected data range greater or equal maximum value, got {data_range} and {x.max()}.'
-    x = x * 255. / data_range
+    x = x / float(data_range) * 255
```
snk4tr, line 54: What is the purpose of the following type conversion: `float(data_range)`? `data_range` (according to the type annotation) is either `int` of `float`. `\` operator always produces `float` result, hence the explicit type conversion is not needed. 
```suggestion
    x = x / data_range * 255
```

Please change this pattern (`float(data_range)`) in all other places if agree.

piq/gmsd.py
```diff
@@ -43,9 +43,11 @@ def gmsd(prediction: torch.Tensor, target: torch.Tensor, reduction: str = 'mean'
         https://arxiv.org/pdf/1308.3052.pdf
     """
 
-    _validate_input(input_tensors=(prediction, target), allow_5d=False, scale_weights=None)
+    _validate_input(
+        input_tensors=(prediction, target), allow_5d=False, scale_weights=None, data_range=data_range)
     prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
 
+    # Rescale to [0, 1] range
```
snk4tr, line 50: This comment is confusing. The operation below does not always scale to the [0, 1] range because a user can provide `data_range` that would scale this particular `prediction` and `target` pair to, let's say, [0, 0.5] range. Please update the description.

piq/gmsd.py
```diff
@@ -179,9 +181,14 @@ def multi_scale_gmsd(prediction: torch.Tensor, target: torch.Tensor, data_range:
     Returns:
         Value of MS-GMSD. 0 <= GMSD loss <= 1.
     """
-    _validate_input(input_tensors=(prediction, target), allow_5d=False, scale_weights=scale_weights)
+    _validate_input(
+        input_tensors=(prediction, target), allow_5d=False, scale_weights=scale_weights, data_range=data_range)
     prediction, target = _adjust_dimensions(input_tensors=(prediction, target))
-
+    
+    # Rescale to [0, 255] range
```
snk4tr, line 188: Same as above. It is not guaranteed that the result will be in the [0, 255] range.

piq/haarpsi.py
```diff
@@ -60,8 +60,8 @@ def haarpsi(x: torch.Tensor, y: torch.Tensor, reduction: str = 'mean',
                          f'Kernel size: {kernel_size}')
 
     # Scale images to [0, 255] range as in the paper
```
snk4tr, line 62: Same as above (about the resulting data range).

piq/ssim.py
```diff
@@ -207,11 +211,14 @@ def multi_scale_ssim(x: torch.Tensor, y: torch.Tensor, kernel_size: int = 11, ke
...
+        input_tensors=(x, y), allow_5d=True, kernel_size=kernel_size,
+        scale_weights=scale_weights, data_range=data_range
+    )
     x, y = _adjust_dimensions(input_tensors=(x, y))
-    if isinstance(x, torch.ByteTensor) or isinstance(y, torch.ByteTensor):
-        x = x.type(torch.float32)
-        y = y.type(torch.float32)
+
+    x = x / float(data_range)
+    y = y / float(data_range)
```
snk4tr, line 221: In lines 54-58 of the same file you first convert tensors to `torch.float32` and then devide by `float(data_range)`. Here the first step is skipped. Please make these two cases consistent.

piq/utils/common.py
```diff
@@ -48,6 +49,9 @@ def _validate_input(
             assert torch.all(tensor >= 0), 'All tensor values should be greater or equal than 0'
         if tensor.dim() == 5:
             assert tensor.size(-1) == 2, f'Expected Complex 5D tensor with (N, C, H, W, 2) size, got {tensor.size()}'
+        if data_range is not None:
+            assert data_range >= tensor.max(), \
+                f'Data range should be greater or equal to maximum tensor value, got {data_range} and {tensor.max()}.'
```
snk4tr, line 54: Nice one! 👍

tests/test_psnr.py
```diff
@@ -16,49 +17,45 @@ def target() -> torch.Tensor:
...
-    psnr(prediction, target, data_range=1.0)
-
-
+def test_psnr(input_tensors: Tuple[torch.Tensor, torch.Tensor], device: str) -> None:
+    prediction, target = input_tensors
+    psnr(prediction.to(device), target.to(device), data_range=1.0)
+
+
+@pytest.mark.parametrize(
+    "data_range", [128, 255],
```
snk4tr, line 26: General remark to all the series of tests: could you add more options of the `data_range` values if it is not too computationally heavy? It would be very interesting to see what happens if `data_range` is 1.0, <1.0 (for instance 0.5) of some weird huge number like 3000.

snk4tr (changes requested):
I like the PR a lot. Only small remarks from my side.

zakajd:
@snk4tr  Ready to be merged

snk4tr (approved):

Issue #188, opened 2020-11-25
Metric: SR-SIM


leihuayi:
**Is your feature request related to a problem? Please describe.**
Spectral Residual based Similarity (SR-SIM) index between two images is the combination of saliency maps made on luminance channel of image and gradient maps (like in FSIM).
*Performance* : Seems to perform slightly better than IW-SSIM according to own paper and HaarPSI paper, and the following studies:
- [A comprehensive evaluation of full-reference image quality assessment algorithms on KADID-10k](https://arxiv.org/pdf/1907.02096), 2019
- [Full-Reference Quality Metric Based on Neural Network to Assess the Visual Quality of Remote Sensing Images](https://www.mdpi.com/2072-4292/12/15/2349/htm#B90-remotesensing-12-02349), 2020

**Describe the solution you'd like**
A pytorch implementation of SR-SIM

**Describe alternatives you've considered**
I have coded a python version out of the matlab code (very similar, except that matlab resize and cv2 resize do not work exactly the same). However I feel it would be more useful to the community if it were part of your repo which is very complete and functional.
I am not very knowledgeable in Pytorch but if you do not have time I can try to implement this feature as contributor.

**Additional context**
Original paper : https://sse.tongji.edu.cn/linzhang/ICIP12/ICIP-SR-SIM.pdf
Authors matlab code : https://github.com/Netflix/vmaf/blob/master/matlab/strred/SR_SIM.m
Python code : https://github.com/leihuayi/FRIQAMetrics/blob/master/friqametrics/srsim.py

PR #202, opened 2021-01-06
Metric: SR-SIM

leihuayi:
Closes [#188](https://github.com/photosynthesis-team/piq/issues/188)

## Proposed Changes

  - Metric SR-SIM in Pytorch
  - Used deprecated functions for fft with 2-dim representation of complex numbers (a,b) because Torch autograd seems not to support newest Torch 1-d representation of complexs (a + ib)
  - Proposed chromatic version of this metric, which does not exists in the original paper. I designed it based on the very strong similarities between SSIM and SR-SIM, it is disabled by default, I can remove it if preferred
  - Added [implementation of matlab-like imresize](https://github.com/thstkdgus35/bicubic_pytorch) (default:bicubic) in piq/functional in order to have results similar to paper
  - Updated doc
  - Tests working with torch>=1.5.0 + torchvision>=0.6.0 up to latest stable versions
  - Suggestion: matlab adds zero padding on even-size filters when running `imfilter(image, filter)`. I manually padded my filter in SR-SIM but it might be useful to directly add the padding in the piq/filters implementations


zakajd:
Hi!
Nice job, thank you for PR. I skimed it through and it looks good. 
I'll review it in details in a few days.

zakajd:
@leihuayi have you tested chromatic version on common MOS databases, such as TID2013, LIVE, KADID10k? 
Does it improve metric performance in terms of PLCC/SRCC?
If not, I can run experiments and post results here for the future comparisons.

leihuayi:
@zakajd no I did not test the chromatic version on databases, I would be very grateful if you could take care of it. Maybe the chromatic constants can be a little tweaked using for instance the constants of VSI, for the moment I just used the constants of SSIM.

zakajd:
Just runed experiments on MOS databases. Here are results:
**TID2013**:
SR-SIM(piq): PLCC 0.7550, SRCC 0.7257,  KRCC 0.5504
SR-SIM(matlab on TID2008): PLCC 0.8866, SRCC 0.8913, KRCC 0.7149 
SR-SIMc(piq): PLCC 0.7800, SRCC 0.7731,  KRCC 0.5901

**KADID10k**
SR-SIM(piq): PLCC 0.5737, SRCC 0.6329,  KRCC 0.4586
SR-SIM([matlab](https://arxiv.org/pdf/1907.02096.pdf)): PLCC 0.639 SRCC 0.839 KRCC 0.652
SR-SIMc(piq): PLCC 0.6306, SRCC 0.6991,  KRCC 0.5147

Chromatic version really does improve the performance 💪 
Also looks like there is some 🐛 in the implementation, performance is worse than MATLAB. Need to investigate it further. 

Commit hash f30c7f27df7804a067d469283c07b93067f8f7a9

leihuayi:
Hi @zakajd thanks a lot for the tests.
I have to admit I have no idea where the discrepancy comes from. After reducing the differences between matlab & python code at each step by comparing the intermediate matrix results, I tested matlab & python code on about 10 paris of images with everytime the same result at 10E-5. The only lead I see is that the default matlab code uses an even-sized kernel for a gaussian, which is pretty strange, I did my best to code a similar behaviour in case of even-sized filters but it did increase a bit the difference between matlab & python results. Did you manually obtain the matlab results you quoted above or did you use the results from the paper ?

zakajd:
Sorry for long answer, was busy with other tasks.
I didn't compute SRCC and KRCC values on TID2013 dataset myself for MATLAB code. I'm using results from 2 papers: [Mean Deviation Similarity Index](https://arxiv.org/pdf/1608.07433.pdf) and [A Comprehensive Performance Evaluation of Image Quality Assessment Algorithms ](https://ieeexplore.ieee.org/abstract/document/8847307). They both independently got similar values of SRCC **~0.8073-8076**  and KRCC **0.6406**. 

What MATLAB version did you use? ❓  I don't have licence for MATLAB so using [octave-online](https://octave-online.net) instead. When computing SR-SIM between `I01.BMP` and `i01_01_05.bmp` I got **0.9583**, not _0.9666_ (as in your tests). 

Spent a couple of hours trying to match results of Octave with PyTorch but failed. Main source of discrepancy comes from wrong interpolation at the very beginning, but neither `torch.nn.functional.interpolate`, neither `piq.functional.resize` match exactly.
Surprisingly, replacing resize by PyTorch `interpolate` function slightly improved the performance: SRCC _0.7257_ -> **0.7337**, KRCC _0.5504_ ->  **0.5657**

leihuayi:
Hi @zakajd ,
no problem for the delay.
I have a student licence so I used Matlab online (which says version 21.0, R2020b (9.9.0)). I originally used octave but somehow I could not make matrix results match after doing a resize (at the beginning of saliency map function), be it using F.interpolate or "matlab-like imresize code". The best result I could get with Octave was using F.interpolate(mode=bicubic, align_corners=False). When I tried using matlab the resize immediately matched using "matlab-like imresize code". Therefore I figured there are some slight differences between Octave and Matlab and therefore if I want to stick as much as possible to the paper I should use the latter one. Like you I don't have the same results using Octave and Matlab when running the original code (respectedly 0.9583 and 0.9666 for i01.bmp image).

denproc:
Performance on KADID10k:
SR-SIM(matlab): SRCC 0.839 KRCC 0.652
SR-SIM (piq): SRCC 0.839 KRCC 0.652
SR-SIMc (piq):  SRCC 0.869 KRCC 0.685

Ready for review.

examples/image_metrics.py
```diff
@@ -130,6 +130,12 @@ def main():
     vsi_loss: torch.Tensor = piq.VSILoss(data_range=1.)(x, y)
     print(f"VSI index: {vsi_index.item():0.4f}, loss: {vsi_loss.item():0.4f}")
 
+    # To compute SR-SIM score as a measure, use lower case function from the library:
+    srsim_index: torch.Tensor = piq.srsim(x, y, data_range=1.)
+    # In order to use VSI as a loss function, use corresponding PyTorch module:
```
zakajd, line 135:
```suggestion
    # In order to use SR-SIM as a loss function, use corresponding PyTorch module:
```

piq/srsim.py
```diff
@@ -0,0 +1,279 @@
...
+
+    eps = torch.finfo(score.dtype).eps
+    result = score.sum(dim=[1, 2, 3]) / (svrs_max.sum(dim=[1, 2, 3]) + eps)
+
+    if reduction == 'none':
+        return result
+
+    return {'mean': result.mean,
+            'sum': result.sum
+            }[reduction](dim=0)
```
zakajd, line 126:
```suggestion
    return _reduce(result, reduction)
```

piq/srsim.py
```diff
@@ -0,0 +1,279 @@
...
+"""
+import functools
+from typing import Union
+
+import torch
+import torch.nn.functional as F
+
+from torch.nn.modules.loss import _Loss
+
+from piq.utils import _validate_input, _version_tuple
```
zakajd, line 15:
```suggestion
from piq.utils import _validate_input, _version_tuple, _reduce
```

zakajd (changes requested):
Great job!
A few minor changes and it can be merged

zakajd (approved):
LGTM 👍🏼

PR #238, opened 2021-03-03
Fix bug in DISTS

zakajd:
Found a bug in DISTS which led to poor performance

## Proposed Changes

  - Add downsampling before score computation to boost performance and reduce memory consumption
  - Adjust tests
  - Update table in README.rst

Performance now **much** closer to the original, but still has some gap. No idea for now what may cause this


denproc (approved):
💪🏻

PR #237, opened 2021-03-01
Get rid of redundant libraries

snk4tr:
Closes #15 

## Proposed Changes

  - Remove `gudhi` and `scipy` from the general `requirements.txt` because they are only used in the `geometry score`, which is itself rarely used
  - In case if the `geometry score` is used and any of required libraries are not installed, throw a warning with instructions on how to installthem
  - Refine tests for the `geometry score` so they do not crash because of changes in the general `requirements.txt`

PR #244, opened 2021-03-31
Exceptional import of scipy in MSID and validation of torch

### LINKED ISSUE #15: Get rid of redundant libraries like scipy
Currently some implementations use libraries for computations on CPU like scipy. 

Refactor the code in such a way that it will work without them.

denproc:
Closes #15 

## Proposed Changes

  - Added exceptional import to MSID as copy is not listed in requirements
  - Validate the `torch` version via tuple instead of strings

### DIFF
--- a/piq/msid.py
+++ b/piq/msid.py
@@ -2,14 +2,13 @@ r"""Implemetation of Multi-scale Evaluation metric, based on paper
  https://arxiv.org/abs/1905.11141 and author's repository https://github.com/xgfs/msid
 """
 from typing import List, Tuple, Optional
+from warnings import warn
 
 import torch
 import numpy as np
 
-from scipy.sparse import lil_matrix, diags, eye
-
 from piq.base import BaseFeatureMetric
-from piq.utils import _validate_input
+from piq.utils import _validate_input, _version_tuple
 
 
 EPSILON = 1e-6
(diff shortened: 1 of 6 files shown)


denproc:
Blocked by #245 #246

piq/msid.py
```diff
@@ -278,6 +279,17 @@ def _msid_descriptor(x: np.ndarray, ts: np.ndarray = np.logspace(-1, 1, 256), k:
...
+    try:
+        import scipy
+    except ImportError:
+        raise ImportError("Scipy is required for computation of the Geometry Score but not installed. "
+                          "Please install scipy using the following command: pip install --user scipy")
+
+    recommended_scipy_version = "1.3.3"
+    if _version_tuple(scipy.__version__) < _version_tuple(recommended_scipy_version):
+        warn(f'Scipy of version {scipy.__version__} is used while version >= {recommended_scipy_version} is '
+             f'recommended. Consider updating scipy to avoid potential problems.')
```
zakajd, line 291: Can we specify exactly which problems may happen when using older versions?

denproc: It was introduced in #237. @zakajd @snk4tr, what was the problem there with older `scipy`?
I remember only slow processing in case of `1.4` version.

snk4tr: Yep, this recommendation was motivated by a long compute time with older versions.

denproc: Resolved

zakajd (approved):
LGTM 👍

PR #246, opened 2021-04-01
Change import FFT according to interface of torch

### LINKED ISSUE #245: Bug: torch.fft functionality
**Describe the bug**
The new version of the `torch==1.8` does not support `torch.fft()` functionality. Now it should be used as `torch.fft.rfft()` since `torch == 1.7.0`.

**To Reproduce**
1. Update pytorch and torchvision
2. Run `pytest test`
3. See error 

**Suggestion**
Upgrade the requirements and change the code correspondently.

denproc:
Closes #245 

## Proposed Changes

  - Added conditional import depending on installed version of `torch`
  - Added CI testing on 2 pytorch versions (`torch=={'1.5.1', '1.8.1'}, torchvision=={'0.6.1', '0,9,1'}`)
  - Changes to be compatible with `torch>=1.5.1`

### DIFF
--- a/piq/brisque.py
+++ b/piq/brisque.py
@@ -53,7 +53,7 @@ def brisque(x: torch.Tensor,
     assert kernel_size % 2 == 1, f'Kernel size must be odd, got [{kernel_size}]'
     _validate_input([x, ], dim_range=(4, 4), data_range=(0, data_range))
 
-    x = x / data_range * 255
+    x = x / float(data_range) * 255
 
     if x.size(1) == 3:
         x = rgb2yiq(x)[:, :1]
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -17,7 +17,7 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        torchvision-version: [0.4.0, 0.9.1]
+        torchvision-version: [0.6.1, 0.9.1]
 
     steps:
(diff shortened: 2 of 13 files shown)


denproc:
In order to resolve blocked merging, the protection of master branch should be changed (required tests). I suggest to do it after review.

piq/brisque.py
```diff
@@ -53,7 +53,7 @@ def brisque(x: torch.Tensor,
     assert kernel_size % 2 == 1, f'Kernel size must be odd, got [{kernel_size}]'
     _validate_input([x, ], dim_range=(4, 4), data_range=(0, data_range))
 
-    x = x / data_range * 255
+    x = x / float(data_range) * 255
```
snk4tr, line 56: Why is this _explicit_ conversion necessary?

snk4tr (commented):
One question from my side. The rest looks good.

piq/brisque.py
denproc: In case of `x` and `data_range` being `torch.uint8` and `int`, the result will be `Byte` for `torch==1.5.1` due to classic division instead of the `true_division`. It will lead to zeroing the input and possible exceptions from pytorch layers, which could be incompatible with `torch.Byte`.

snk4tr: Agh, I see. This is unfortunate because from one hand we don't check user's inputs but from the other we require `data_range` to be `int` of `float`. I think this is a good compromise though.

snk4tr (approved):

PR #259, opened 2021-07-27
Layers/Weights Bug in Content Loss 

### LINKED ISSUE #258: Potential Bug in Content Loss
**Describe the bug**
If the length of `layers` mismatches `weights`, the content loss uses only the first layer **silently**.

**To Reproduce**

```python
import torch
import piq

img = torch.ones((1, 3, 256, 256))
ref_img = torch.zeros((1, 3, 256, 256))

piq.ContentLoss(layers=['relu1_2', 'relu2_2'])(img, ref_img)
# tensor(43.9598)
piq.ContentLoss(layers=['relu1_2'])(img, ref_img)
# tensor(43.9598)
piq.ContentLoss(layers=['relu1_2', 'relu2_2'], weights=[1.0, 1.0])
# tensor(214.3362)
```

**Expected behavior**
raise warnings or errors if the length of `layers` mismatches `weights`.

snk4tr:
Closes #258

## Proposed Changes

  - Fix bug with layers/weighs length in the ContentLoss (if number of layers > number of weights => extra layers are skipped => wrong results) via addition of an assertion. Now mismatch will cause an error.
  - Add required tests 
  - Fix already existing tests that start failing

### DIFF
--- a/piq/perceptual.py
+++ b/piq/perceptual.py
@@ -139,8 +139,6 @@ class ContentLoss(_Loss):
             else:
                 raise ValueError("Unknown feature extractor")
 
-        self.weights = [torch.tensor(w) for w in weights]
-
         if replace_pooling:
             self.model = self.replace_pooling(self.model)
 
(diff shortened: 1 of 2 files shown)


snk4tr:
Ready for review. Btw I have no idea why coverage of `gs.py` decreased since I only fixed `perceptual.py` and the corresponding test file.

piq/perceptual.py
```diff
@@ -133,6 +139,8 @@ def __init__(self, feature_extractor: Union[str, torch.nn.Module] = "vgg16", lay
             else:
                 raise ValueError("Unknown feature extractor")
 
+        self.weights = [torch.tensor(w) for w in weights]
```
denproc, line 142: Duplicate of line 156

denproc (changes requested):
Minor issue with doubled weight transformation to list of tensors. We can keep any one, but we must not forget about check if `w` is a tensor to avoid UserWarning.

snk4tr:
@denproc ready now.

PR #263, opened 2021-08-01
pypi-friendly readme

denproc:
Closes #256 

## Proposed Changes
- Added proper readme format for long description in `setup.py`
- Changed relative link to the absolute for logo image

@snk4tr, please remove 0.5.5.1 and 0.5.5.2 from pypi, which were used for testing. I do not have access. The 0.5.5 release could be removed from PYPI and upgraded with this PR as well since there were no changes in master branch.

Ready for review.

PR #264, opened 2021-08-01
Eliminate UserWarning

### LINKED ISSUE #255: UserWarning: incorrect tensor copy
**Describe the bug**

```
...\.venv\lib\site-packages\piq\perceptual.py:139: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  self.weights = [torch.tensor(w) for w in weights]
```

**To Reproduce**
This happens with perceptual losses, maybe some others too. 

**Expected behavior**
No warning.

denproc:
Closes #255 

## Proposed Changes
- Added type check to eliminate double wrapping in case of input list of tensors in `perceptual.py`

Merge after #263.

### DIFF
--- a/piq/perceptual.py
+++ b/piq/perceptual.py
@@ -145,7 +145,7 @@ class ContentLoss(_Loss):
             "mae": nn.L1Loss,
         }[distance](reduction='none')
 
-        self.weights = [torch.tensor(w) for w in weights]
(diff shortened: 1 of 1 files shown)


snk4tr (approved):
@denproc feel free to merge when ready 😄

PR #270, opened 2021-09-05
Extend benchmarks with computation of scores for NR and DB metrics

### LINKED ISSUE #265: Extend results benchmark
Currently, the [results benchmark](https://github.com/photosynthesis-team/piq/blob/master/tests/results_benchmark.py) does not work with no-reference and distribution-based metrics. Extend the benchmark script to make fair comparison between various categories of metrics possible.

Extend the `eval_metric` function to make it work with different categories of metrics.
Create a `Metric` dataclass to make logic simpler (reduce number of parameters).

Introduce independent methods for different categories of metrics.

snk4tr:
Closes #265 

## Proposed Changes

  - Add computation of correlation scores for no-reference metrics on all datasets
  - Add computation of correlation scores for distribution-based (DB) metrics on all datasets
  - Add description of how DB metrics are computed + reference to the paper on IQA for MRI


snk4tr:
Ready for review.

zakajd (changes requested):
I like the implementation, thanks @snk4tr 
Let's add benchmarks results into README table and also some explanation on how values for distribution-based metrics are computed.

snk4tr:
Had to force push to trigger CI, which was stuck otherwise. See [stackoverflow issue](https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported) on that.

tests/results_benchmark.py
```diff
@@ -73,27 +103,31 @@ def __init__(self, root: Path = "datasets/tid2013"):
         self.df = df[["dist_img", 'ref_img', 'score']]
         self.root = root
 
-    def __getitem__(self, index):
+    def __getitem__(self, index) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         x_path = self.root / self.df.iloc[index][0]
         y_path = self.root / self.df.iloc[index][1]
         score = self.scores[index]
 
         # Load image and ref. Convert to tensor and [0, 1] range
```
zakajd, line 111:
```suggestion
        # Load image and ref, convert to tensor.
```

tests/results_benchmark.py
```diff
@@ -137,62 +171,164 @@ def __init__(self, root: Path = Path("data/raw/pipal")):
...
         distorted_images, reference_images = distorted_images.to(device), reference_images.to(device)
         gt_scores.append(scores.cpu())
 
-        # Full Reference methods
-        metric_score = metric(distorted_images, reference_images).cpu()
+        metric_score: torch.Tensor = \
+            compute_function(metric.functor, distorted_images, reference_images, device, feature_extractor)
+
+        if np.isnan(metric_score):
+            metric_score = torch.Tensor([0])
```
zakajd, line 202: If metric returns NaN shouldn't we raise an exception? 
What a valid cases of NaN result that can be replaced with 0?
Why 0, some metrics should be minimised, some other - maximised, this can create confusion.

tests/results_benchmark.py
```diff
@@ -137,62 +171,164 @@ def __init__(self, root: Path = Path("data/raw/pipal")):
...
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Benchmark PIQ metrics")
 
-    # General
     parser.add_argument("--dataset", type=str, help="Dataset name", choices=list(DATASETS.keys()))
     parser.add_argument("--path", type=Path, help="Path to dataset")
     parser.add_argument('--metrics', nargs='+', default=[], help='Metrics to benchmark', choices=list(METRICS.keys()))
-    parser.add_argument('--batch_size', type=int, default=1, help='Batch size')
```
zakajd, line 187: Why you deleted `batch_size` option? It can be useful in case if dataset consists of same sized images.

zakajd (changes requested):

tests/results_benchmark.py
snk4tr: The script is meant to compute correlation values between MOSs and metrics' scores. If `batch_size > 1`, the script computed a single metric score for `batch_size` number of images. In this case:
1. Miscorrespondance between number of MOSs and number of metrics' scores causes a crush
2. We may try to fix it by computing some aggregates of MOSs, which I decided not to do because the usage could be error-prone

tests/results_benchmark.py
snk4tr: I am observing problems with NaN values for some DB metrics (e.g., IS) on KADID10k dataset. I am not sure about the reason of such an output at this point in time. I do not consider this case to be valid.

tests/results_benchmark.py
snk4tr: Nice catch. I removed it when I was working with medical images of various sizes. Returning back

tests/results_benchmark.py
snk4tr: Removing this line (especially considering that it does not work with `batch_size > 1`). I think that we need to investigate the reasons of this behaviour in a different PR

tests/results_benchmark.py
snk4tr: Check out the issue #309

tests/results_benchmark.py
snk4tr: Check out the issue https://github.com/photosynthesis-team/piq/issues/309

tests/results_benchmark.py
```diff
@@ -137,62 +176,170 @@ def __init__(self, root: Path = Path("data/raw/pipal")):
...
-    for name in metrics:
-        gt_scores, metric_scores = eval_metric(loader, METRICS[name], device=device)
-        print(f"{name}: SRCC {abs(spearmanr(gt_scores, metric_scores)[0]):0.3f}",
+    if metrics[0] == 'all':
+        metrics = list(chain(*METRIC_CATEGORIES.values()))
+
+    for metric_name in metrics:
+        metric: Metric = METRICS[metric_name]
+        gt_scores, metric_scores = eval_metric(loader, metric, device=device, feature_extractor=feature_extractor)
+        print(f"{metric_name}: SRCC {abs(spearmanr(gt_scores, metric_scores)[0]):0.3f}",
```
denproc, line 319: Good to have feature but not necessary one. Would it be possible to log the results into a file to minimise risks of losing outputs in stdout?

denproc (changes requested):
Great work. Just some minor changes.

Line 200: I would change it to "Feature Based"

tests/results_benchmark.py
```diff
@@ -137,62 +176,170 @@ def __init__(self, root: Path = Path("data/raw/pipal")):
...
+    return distorted_features, reference_features
+
+
+def normalize_tensor(tensor: torch.Tensor) -> torch.Tensor:
+    """ Map tensor values to [0, 1] """
+    return (tensor - tensor.min()) / (tensor.max() - tensor.min())
+
+
+def compute_distribution_based(metric_functor: Callable, distorted_images: torch.Tensor,
+                               reference_images: torch.Tensor, device: str, feature_extractor_name: str) -> np.ndarray:
```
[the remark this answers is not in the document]
snk4tr: Indeed

tests/results_benchmark.py
```diff
@@ -137,62 +176,170 @@ def __init__(self, root: Path = Path("data/raw/pipal")):
 }
 
 
-def eval_metric(loader: torch.utils.data.DataLoader, metric: Callable, device: str) -> List:
+def eval_metric(loader: DataLoader, metric: Metric, device: str, feature_extractor: str) \
+        -> Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
snk4tr: Unfortunately, `torch.Tensor` cannot be used with `spearmanr` and `kendalltau`, explicit type conversion is required. I incorporated your suggestion with later type conversion.

tests/results_benchmark.py
snk4tr: I would keep the code minimal. In practice, logging into a file is as simple as running the script with the following line:  
`$ python results_benchmark.py >> results.txt`

snk4tr:
@denproc @zakajd ready for review.

tests/results_benchmark.py
```diff
@@ -73,27 +109,40 @@ def __init__(self, root: Path = "datasets/tid2013"):
...
+        x: image with some kind of distortion in [0, 1] range
+        y: image without distortion in [0, 1] range
+        score: MOS score for this pair of images
+    """
     _filename = "dmos.csv"
 
     def __init__(self, root: Path = "datasets/kadid10k"):
-        assert root.exists(),\
-            "You need to download KADID10K dataset first. Check http://database.mmsp-kn.de/kadid-10k-database.html"
+        assert root.exists(), \
```
denproc, line 142: Missing "-> None"

denproc (commented):
LGTM, just a minor change to make documentation consistent

tests/results_benchmark.py
denproc: Fixed with the following commit

denproc (dismissed):

PR #278, opened 2021-11-08
speed up CI

bonlime:
## Proposed Changes

Speed up CI by caching python + installed dependencies

upd. it seems that your cache is too large. 1+ Gb for each element in the matrix. Github gives you only 5 Gb in cache, so only 4/6 runs could be cached, others would still require install. This still would save ~8 mins each time, so merge if it looks good to you


bonlime:
why did you close it? it wouldn't hurt

zakajd:
@bonlime Yesterday on a monthly call we decided to skip this enhancement.

Issue #288, opened 2022-02-02
Compute error for `PSNR` and `SSIM`


leftthomas:
Convert RGB image to YCbCr and take luminance should be `Y = 0.256789 R + 0.504129 G +0.097906 B`, not
https://github.com/photosynthesis-team/piq/blob/e8f6513afa8f9138bcd19c0928763cc18bf7e060/piq/psnr.py#L39

As for `SSIM`, to make sure the results can be same with [Matlab Code](https://github.com/hongwang01/RCDNet/tree/master/Performance_evaluation), it must use `DoubleTensor`, and these three lines should change:
https://github.com/photosynthesis-team/piq/blob/e8f6513afa8f9138bcd19c0928763cc18bf7e060/piq/ssim.py#L55
https://github.com/photosynthesis-team/piq/blob/e8f6513afa8f9138bcd19c0928763cc18bf7e060/piq/ssim.py#L56
https://github.com/photosynthesis-team/piq/blob/e8f6513afa8f9138bcd19c0928763cc18bf7e060/piq/functional/filters.py#L35

Please refer to [my implement](https://github.com/leftthomas/RCDNet/blob/master/utils.py) of `SSIM` and `PSNR`.

zakajd:
Hi @leftthomas thanks for raising the issue.
I can't yet agree with your point on PSNR conversion formula. Can you please provide a reference that mentions `Y = 0.256789 R + 0.504129 G +0.097906 B` values? Wikipedia pages on both [YUV](https://en.wikipedia.org/wiki/YUV) and [YCbCr](https://en.wikipedia.org/wiki/YCbCr) mention values we currently use with reference to ITU-R BT.601 standard. 

With regard to DoubleTensor — how much is the absolute difference between 2 methods using single and double precisions? 
BTW link to your implementation shows 404 error.

leftthomas:
@zakajd I have make the repo of my implementation public, and you could refer to. 

With regard to DoubleTensor — I have tested with `Rain1400` in terms of `SSIM`, which is a Derain dataset, the absolute difference between 2 methods using single and double precisions is 0.0003, this value can not be ignored for `Derain task`, and I think you could refactor the API as like this:
``
coords = torch.arange(kernel_size, dtype=input.dtype, device=input.device)
``
this could compatible with both `FloatTensor` and `DoubleTensor`, according to input data.

[...]

PR #292, opened 2022-02-06
Create variables on correct devices and with correct types

zakajd:
Related to #288 

## Proposed Changes

  - Propagate input dtype though whole computation pipeline and cast intermediate kernels / filters to it. 
  - Add tests to verify that for float32 and float64. float16 is not included, because not all operations are defined for CPU.


denproc (changes requested):
Great fix!
- For BRISQUE, '_natural_scene_statistics' uses gaussian kernel (line 178), which is generated as 'torch.FloatTensor' losing the precision for 'torch.DoubleTensor' input. I suggest to propagate type into the kernel generation or use 'torch.float64' as default for generating kernels.
- For DSS, I suggest type assignment in line 90 to avoid implicit type upgrade in following loop. Function '_dct_matrix' and ‘gaussian_fliter’ produce 'torch.FloatTensor's losing the precision if 'torch.DoubleTensor's are used as the input. Therefore, I suggest, to add type propagation inside.
- For FSIM, '_lowpassfilter' amd ‘scharr_filter’ are 'torch.FloatTensor'. Similar suggestion to the previous. Propagate types or generate 'torch.float64' as default with following reduction to 'torch.float32'. Same for ‘_construct_filters’ in lines 154-158.
- For MS-SSIM and SSIM, we use gaussian filter, generated using ‘torch.FloatTensor’ in line 73 (MS-ssim) and in line 64 (SSIM).
- For VSI, ‘scharr_filter’ and ‘_log_gabor’ are generated as ‘torch.FloatTensor’.
- For other functionality, that wasn't changed with presented commits, could contain same issues.

Concluding on the above, ‘functional/filters.py’ could be upgraded to ‘torch.float64’ as default precision, which could be reduces after.

zakajd:
As discussed with Denis and Sergei, we won't create kernels as `torch.DoubleTensor` by default.

**Reasoning:**
 It's much more common to have single precision float tensors as inputs. So having kernels in float64 will create _a lot_ of casting back to float32 every time metrics are computed.

This PR adds support for float64 input inference and slight loss of precision due to kernel's dtype is not considered important.

@denproc please review again and merge.

denproc (approved):
LTGM

PR #305, opened 2022-02-25
Fix Exceptions for DSS, GMSD, Haarpsi, MS-SSIM, Perceptual, SRSIM, SSIM, VIF

### LINKED ISSUE #304: ValueError instead of AssertionError for DSS, GMSD, Haarpsi, MS-SSIM, Perceptual, SRSIM, SSIM, VIF
DSS, GMSD, Haarpsi, MS-SSIM, Perceptual, SRSIM, SSIM, VIF metrics and loss function classes use `ValueError` exception instead of `AssertionError`.

**Expected behavior**
Check size of the images or features with `assert.`

**Additional context**
In context of previous discussion to allow user to turn off validation with `python -O` command to boost processing time.

denproc:
Closes #304 

## Proposed Changes
- Changed `ValueError` to `AssertionError` for DSS, GMSD, Haarpsi, MS-SSIM, Perceptual, SRSIM, SSIM, VIF
- Changed comments for Exception in accordance with validation in case of DSS

### DIFF
--- a/piq/kid.py
+++ b/piq/kid.py
@@ -29,13 +29,11 @@ def _polynomial_kernel(X: torch.Tensor, Y: torch.Tensor = None, degree: int = 3,
     if Y is None:
         Y = X
 
-    if X.dim() != 2 or Y.dim() != 2:
-        raise ValueError('Incompatible dimension for X and Y matrices: '
-                         'X.dim() == {} while Y.dim() == {}'.format(X.dim(), Y.dim()))
+    assert X.dim() == 2 and Y.dim() == 2,\
+        f'Incompatible dimension for X and Y matrices: X.dim() = {X.dim()}, while Y.dim() = {Y.dim()}'
 
-    if X.size(1) != Y.size(1):
-        raise ValueError('Incompatible dimension for X and Y matrices: '
-                         'X.size(1) == {} while Y.size(1) == {}'.format(X.size(1), Y.size(1)))
+    assert X.size(1) == Y.size(1), \
+        f'Incompatible dimension for X and Y matrices: X.size(1) == {X.size(1)} while Y.size(1) == {Y.size(1)}'
 
     if gamma is None:
         gamma = 1.0 / X.size(1)
--- a/piq/kid.py
+++ b/piq/kid.py
@@ -29,12 +29,6 @@ def _polynomial_kernel(X: torch.Tensor, Y: torch.Tensor = None, degree: int = 3,
     if Y is None:
         Y = X
 
-    assert X.dim() == 2 and Y.dim() == 2,\
-        f'Incompatible dimension for X and Y matrices: X.dim() = {X.dim()}, while Y.dim() = {Y.dim()}'
-
-    assert X.size(1) == Y.size(1), \
-        f'Incompatible dimension for X and Y matrices: X.size(1) == {X.size(1)} while Y.size(1) == {Y.size(1)}'
-
     if gamma is None:
         gamma = 1.0 / X.size(1)
 
--- a/piq/ssim.py
+++ b/piq/ssim.py
@@ -186,9 +186,8 @@ def _ssim_per_channel(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor,
     Returns:
         Full Value of Structural Similarity (SSIM) index.
     """
-    if x.size(-1) < kernel.size(-1) or x.size(-2) < kernel.size(-2):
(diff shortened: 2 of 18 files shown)


piq/ssim.py
```diff
@@ -186,9 +186,8 @@ def _ssim_per_channel(x: torch.Tensor, y: torch.Tensor, kernel: torch.Tensor,
     Returns:
         Full Value of Structural Similarity (SSIM) index.
     """
-    if x.size(-1) < kernel.size(-1) or x.size(-2) < kernel.size(-2):
-        raise ValueError(f'Kernel size can\'t be greater than actual input size. Input size: {x.size()}. '
-                         f'Kernel size: {kernel.size()}')
+    assert x.size(-1) >= kernel.size(-1) and x.size(-2) >= kernel.size(-2), \
+        f'Kernel size can\'t be greater than actual input size. Input size: {x.size()}. Kernel size: {kernel.size()}'
```
snk4tr, line 190: I think this one should remain an exception. Disabling it with python -O makes no sense, the code will fail later anyway. What we want to do with rising this exception is to make it fail early and save compute time.

piq/vif.py
```diff
@@ -46,8 +46,8 @@ def vif_p(x: torch.Tensor, y: torch.Tensor, sigma_n_sq: float = 2.0,
     _validate_input([x, y], dim_range=(4, 4), data_range=(0, data_range))
 
     min_size = 41
-    if x.size(-1) < min_size or x.size(-2) < min_size:
-        raise ValueError(f'Invalid size of the input images, expected at least {min_size}x{min_size}.')
+    assert x.size(-1) >= min_size and x.size(-2) >= min_size, \
+        f'Invalid size of the input images, expected at least {min_size}x{min_size}.'
```
snk4tr, line 51: I think this one should remain an exception. Disabling it with python -O makes no sense, the code will fail later anyway. What we want to do with rising this exception is to make it fail early and save compute time.

snk4tr (changes requested):

snk4tr:
@denproc thanks for the PR. As you see from the review, I think that several changes that you proposed are not actually required. Let me know what you think.
There are several nice catches though 👍 .

piq/kid.py
```diff
@@ -29,13 +29,11 @@ def _polynomial_kernel(X: torch.Tensor, Y: torch.Tensor = None, degree: int = 3,
...
-        raise ValueError('Incompatible dimension for X and Y matrices: '
-                         'X.dim() == {} while Y.dim() == {}'.format(X.dim(), Y.dim()))
+    assert X.dim() == 2 and Y.dim() == 2,\
+        f'Incompatible dimension for X and Y matrices: X.dim() = {X.dim()}, while Y.dim() = {Y.dim()}'
 
-    if X.size(1) != Y.size(1):
-        raise ValueError('Incompatible dimension for X and Y matrices: '
-                         'X.size(1) == {} while Y.size(1) == {}'.format(X.size(1), Y.size(1)))
+    assert X.size(1) == Y.size(1), \
+        f'Incompatible dimension for X and Y matrices: X.size(1) == {X.size(1)} while Y.size(1) == {Y.size(1)}'
```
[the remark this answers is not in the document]
denproc: Indeed, matrix multiplication at line 41 fails in case of different size of the matrices. I would suggest to keep our validation due to better explanation of error compared to exceptions of `torch.mm`.
Tensor with more or less than 2 dimensions:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: mat2 must be a matrix
```
Tensors that can't be multiplied due to their shapes:
```
>>> torch.mm(a,b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: mat1 and mat2 shapes cannot be multiplied (2x3 and 2x3)
```

piq/vif.py
denproc: Let's discuss it in this tread :).

We do use exceptions to raise error at earliest stage. However, we decided to pursue to approach with assertion errors instead of the usual exceptions to allow "pro" users turn off validation for faster processing. As we stated in readme, `-O` option could be used at user's risk with less information for debugging. As a result, we shouldn't worry about detecting the error at earliest stage with `-O` flag used.

Moreover, `kernel_size` parameter verification is an addition to input tensor validation `_validate_input`, which uses `assert`s.

Due to these reasons, I would suggest to use `assert` for validation to support coherence between other validations and our description of `python -O` option.
@snk4tr, please, let me know your feedback.

PR #313, opened 2022-04-18
Change BRISQUE interpolation to match MATLAB performance

### LINKED ISSUE #312: Add comparison of BRISQUE result with reference value
**Describe the bug**
Current implementation produces values that are different from the ones mentioned in reference paper. 
Need to investigate reasons for such behaviour. 

**Expected behavior**
Values obtained on TID2013, KADID10k, PIPAL match with the ones in other papers.

denproc:
Closes #312 

## Proposed Changes
- changed interpolation into Matlab like resize
- changed precision depending on the input for gaussian filter
- updated tests with true scores by computed by original Matlab code with double precision


snk4tr (approved):
GJ @denproc 👍

PR #333, opened 2023-01-27
Update base.py

### LINKED ISSUE #332: Implementation of `gradient_map` does not support >1 channels
**Describe the bug**
According to documentation this function should work for images with arbitrary number of channels, but current implementation fails for C<> 1.

Also documentation says that expected kernel shape is [n, k, k] while in reality we always pass [n, 1, k, k] because it doesn't work otherwise due to the same error in implementation 

**To Reproduce**
```
x = torch.randn(4, 3, 16, 16)
kernels = torch.randn(2, 1, 3, 3) # should be torch.randn(2, 3, 3) 

output = gradient_map(x, kernels)
```

**Expected behavior**
No errors, proper computation. 

**Additional context**
Not sure yet, but probably this function also doesn't support kN != 2. Will check while fixing

jzakirov:
Closes #332 

## Proposed Changes

  - [x] Fixed gradient map computation
  - [ ] Updated all metrics to use proper kernel shapes (optional, function supports both variants now).

### DIFF
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -45,17 +45,27 @@ def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float, al
 
 def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
     r""" Compute gradient map for a given tensor and stack of kernels.
-
     Args:
         x: Tensor with shape (N, C, H, W).
-        kernels: Stack of tensors for gradient computation with shape (k_N, k_H, k_W)
+        kernels: Stack of tensors for gradient computation with shape (k_N, k_H, k_W) or (k_N, 1, k_H, k_W)
     Returns:
         Gradients of x per-channel with shape (N, C, H, W)
     """
     padding = kernels.size(-1) // 2
-    grads = torch.nn.functional.conv2d(x, kernels.to(x), padding=padding)
+    N, C, H, W = x.shape
 
-    return torch.sqrt(torch.sum(grads ** 2, dim=-3, keepdim=True))
+    # Expand kernel if this is not done already and repeat to match number of groups
+    if kernels.dim() != 4:
+        kernels = kernels.unsqueeze(1)
+
+    if C > 1:
+        kernels = kernels.repeat(C, 1, 1, 1)
+        
+    # Process each channel separately using group convolution.
+    grads = torch.nn.functional.conv2d(x, kernels.to(x), groups=C, padding=padding)
+
+    # Create a per-channel view, compute square of grads and return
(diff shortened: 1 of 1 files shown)


piq/functional/base.py
```diff
@@ -49,17 +49,27 @@ def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float, al
 
 def gradient_map(x: torch.Tensor, kernels: torch.Tensor) -> torch.Tensor:
     r""" Compute gradient map for a given tensor and stack of kernels.
-
     Args:
         x: Tensor with shape (N, C, H, W).
-        kernels: Stack of tensors for gradient computation with shape (k_N, k_H, k_W)
+        kernels: Stack of tensors for gradient computation with shape (k_N, k_H, k_W) or (k_N, 1, k_H, k_W)
```
denproc, line 54:
```suggestion
        kernels: Stack of tensors for gradient computation with shape (k_C_out, k_C_in, k_H, k_W). k_C_in equals 1.
```

piq/functional/base.py
```diff
@@ -49,17 +49,27 @@ def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float, al
...
         Gradients of x per-channel with shape (N, C, H, W)
     """
     padding = kernels.size(-1) // 2
-    grads = torch.nn.functional.conv2d(x, kernels, padding=padding)
+    N, C, H, W = x.shape
+
+    # Expand kernel if this is not done already and repeat to match number of groups
+    if kernels.dim() != 4:
+        kernels = kernels.unsqueeze(1)
+
```
denproc, line 64:
```suggestion
    assert kernels.dim() == 4, f'Expected 4D kernel, got {kernels.dim()}D tensor '
    assert kernels.size(1) == 1, f'Expected dimension size of kernel to be equal one for input number of channels, got kernel {kernel.size()} '
    assert kernels.size(-1) == kernel_size(-2), f'Expected squared kernel along coast two dimensions, got {kernel.size()}'

    padding = kernels.size(-1) // 2
    N, C, H, W = x.shape
```

denproc (changes requested):
Hi @zakajd,

As you mentioned in #332, we pass the 4D tensor, not 3D. I would suggest to update the doc string instead of expanding the function to a non standard kernel sizes, which could require additional testing. Edge cases: `kernels.dim() = 5  # or 2`.

I checked the library for every metric that uses `gradient_map`. Usually, we consider only luminance channel to be provided for the function. I assume that there are other cases outside the library, where it could be helpful to compute gradient map per channel, i.e. using the `gradient_map` as standalone function. 
If we follow this path, let's add `assert` on kernel size and tests to cover the features of the function.

Issue #227, opened 2021-02-04
Optimization: Cash kernels and move creation to correct device


zakajd:
**Is your feature request related to a problem? Please describe.**
Most metrics use some kind of kernels for extraction of image features.
Now those kernels are created each time when metric is called. That's slow and redundant when the function is called hundreds of times.

Another problem related to performance is creation of temporal tensors first on CPU and later moving them to GPU if needed (`.to(x.device)`). 

**Describe the solution you'd like**
1. Add support for cashing kernels between runs. Add `kernel` param to functional API. Create and store kernel in class API
2. Replace `.to(x.device)` and simmular calls to explicit creation of tensor on target device. Like `torch.ones(N, N, device=x.device)`.

**Additional context**
For metrics that use more than one kernel implementation details can be discussed.

denproc:
Hi @zakajd,

The second part is partially resolved in #334.

While the proposed changes cover the tensor creation related to filters and colour conversions, some parts of the library could still use `.to` (`piq/base.py`,  `fid.py`, `perceptual.py, `pieapp.py` and `piq/functional/resize.py`).

PR #334, opened 2023-02-05
FSIM Optimisation to Reduce CPU Usage

### LINKED ISSUE #266: High CPU usage for FSIM and FSIMc
**Describe the bug**
When computing FSIM and FSIMc with device set to CUDA computation is still heavily CPU bounded. 
See screenshot of `htop`

**To Reproduce**
1. Launch benchmark code `python3 tests/results_benchmark.py --dataset pipal --metrics FSIM FSIMc --path /data/pipal --batch_size 16`

**Screenshots**
Screenshot 2021-09-02 at 19 48 21

**Additional context**
Related to #227

denproc:
Closes #266 
Addresses the second part of #227 

## Proposed Changes
- FSIM: got rid of a duplicate for filter generation `_construct_filters`
- FSIM: removed the duplicate of mesh grid generation performed for `_lowpassfilter`. The filter is computed explicitly.
- FSIM: minor optimisation of tensor generations across the `fsim` to avoid transfer between devices
- base.py: mesh grid creation (`get_meshgrid`) is performed straight on target device and data type according to input tensor without redundant transfers between devices
- base.py: removed redundant `to()` operation in `gradient_map`
- filters: filters are generated straight on target device and data type according to input tensor without redundant transfers between devices
- colour_conversion: moved tensor generation to the target device instead of transfer between devices
- Updated `.gitignore`
- minor change in benchmark script to allow more flexible choice for GPU device
- minor fixes across the code to match the changes for colour conversion and filter generation

## Outcome of the Proposed Changes

- Reduced cpu usage with `cuda`-enabled computations for FSIM and FSIMc (visual comparison is atttached)
- CPU time was reduced by 6 sec according to PyTorch profiler

## Note
Other places to consider for additional assessment of leaks
- `piq/base.py`,  `fid.py`, `perceptual.py` and `pieapp.py` use `.to()` for a various transfers. They are out of scope for current PR due to a different purpose of the transfer between devices/types.
- custom made `piq/functional/resize.py` could be refactored as well to get rid of redundant transfers. Added to #227 
- `vsi.py` has redundant mesh grid creation. Opened #335 

# Appendix 
## Visual Comparison of CPU usage
`htop` following #266 

### Before
Branch: master
Screenshot 2023-02-05 at 00 42 23

### After
Branch: bug/cpu_usage
Screenshot 2023-02-05 at 00 38 06

## PyTorch Profiler Results. Top-8 operations.
```
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg    # of Calls  
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  
                                        model_inference        16.38%        8.624s       100.00%       52.639s       52.639s             1  
                                               aten::to         0.07%      37.231ms        68.84%       36.234s       1.020ms         35525  
                                         aten::_to_copy         0.14%      72.795ms        68.76%       36.197s       4.993ms          7250  
                                            aten::copy_         0.34%     177.512ms        61.16%       32.196s       2.019ms         15950  
                                  cudaStreamSynchronize        40.22%       21.173s        40.22%       21.173s       1.718ms         12325  
                                        cudaMemcpyAsync        20.68%       10.883s        20.68%       10.883s     883.015us         12325  
                                    aten::empty_strided         0.17%      88.161ms         7.78%        4.093s     564.510us          7250  
                                             cudaMalloc         7.62%        4.009s         7.62%        4.009s     129.312ms            31  
```

### DIFF
--- a/piq/dss.py
+++ b/piq/dss.py
@@ -174,7 +174,7 @@ def _dct_matrix(size: int, device: Optional[str] = None, dtype: Optional[type] =
     p = torch.arange(1, size, device=device, dtype=dtype).reshape((size - 1, 1))
     q = torch.arange(1, 2 * size, 2, device=device, dtype=dtype)
     return torch.cat((
-        math.sqrt(1 / size) * torch.ones((1, size)),
+        math.sqrt(1 / size) * torch.ones((1, size), device=device, dtype=dtype),
         math.sqrt(2 / size) * torch.cos(math.pi / (2 * size) * p * q)), 0)
 
 
--- a/piq/dss.py
+++ b/piq/dss.py
@@ -135,8 +135,8 @@ def _subband_similarity(x: torch.Tensor, y: torch.Tensor, first_term: bool,
     c = dc_coeff if first_term else ac_coeff
 
     # Compute local variance
-    kernel = gaussian_filter(kernel_size=kernel_size, sigma=sigma)
-    kernel = kernel.view(1, 1, kernel_size, kernel_size).to(x)
+    kernel = gaussian_filter(kernel_size=kernel_size, sigma=sigma, dtype=x.dtype, device=x.device)
+    kernel = kernel.view(1, 1, kernel_size, kernel_size)
     mu_x = F.conv2d(x, kernel, padding=kernel_size // 2)
     mu_y = F.conv2d(y, kernel, padding=kernel_size // 2)
 
--- a/piq/dss.py
+++ b/piq/dss.py
@@ -12,7 +12,7 @@ import torch
 
 import torch.nn.functional as F
 
-from typing import Union
+from typing import Union, Optional
 from torch.nn.modules.loss import _Loss
 
 from piq.utils import _validate_input, _reduce
@@ -162,7 +162,7 @@ def _subband_similarity(x: torch.Tensor, y: torch.Tensor, first_term: bool,
     return similarity
 
 
-def _dct_matrix(size: int, dtype: type = None, device: str = None) -> torch.Tensor:
+def _dct_matrix(size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r""" Computes the matrix coefficients for DCT transform using the following formula:
     https://fr.mathworks.com/help/images/discrete-cosine-transform.html
 
--- a/piq/dss.py
+++ b/piq/dss.py
@@ -84,7 +84,7 @@ def dss(x: torch.Tensor, y: torch.Tensor, reduction: str = 'mean',
     dct_y = _dct_decomp(y_lum, dct_size)
 
     # Create a Gaussian window that will be used to weight subbands scores
-    coords = torch.arange(1, dct_size + 1).to(x)
+    coords = torch.arange(1, dct_size + 1, dtype=x.dtype, device=x.device)
     weight = (coords - 0.5) ** 2
     weight = (- (weight.unsqueeze(0) + weight.unsqueeze(1)) / (2 * sigma_weight ** 2)).exp()
 
@@ -162,15 +162,17 @@ def _subband_similarity(x: torch.Tensor, y: torch.Tensor, first_term: bool,
     return similarity
 
 
-def _dct_matrix(size: int) -> torch.Tensor:
+def _dct_matrix(size: int, dtype: type = None, device: str = None) -> torch.Tensor:
     r""" Computes the matrix coefficients for DCT transform using the following formula:
     https://fr.mathworks.com/help/images/discrete-cosine-transform.html
 
     Args:
-       size : size of DCT matrix to create.  (`size`, `size`)
+        size : size of DCT matrix to create.  (`size`, `size`)
+        device: target device for generation
+        dtype: target data type for generation
     """
-    p = torch.arange(1, size).reshape((size - 1, 1))
-    q = torch.arange(1, 2 * size, 2)
+    p = torch.arange(1, size, device=device, dtype=dtype).reshape((size - 1, 1))
+    q = torch.arange(1, 2 * size, 2, device=device, dtype=dtype)
     return torch.cat((
         math.sqrt(1 / size) * torch.ones((1, size)),
         math.sqrt(2 / size) * torch.cos(math.pi / (2 * size) * p * q)), 0)
@@ -196,7 +198,7 @@ def _dct_decomp(x: torch.Tensor, dct_size: int = 8) -> torch.Tensor:
     blocks = blocks.view(bs, 1, -1, dct_size, dct_size)  # shape (bs, 1, block_num, N, N)
 
     # apply DCT transform
-    coeffs = _dct_matrix(dct_size).to(x)
+    coeffs = _dct_matrix(dct_size, device=x.device, dtype=x.dtype)
 
     blocks = coeffs @ blocks @ coeffs.t()  # @ does operation on last 2 channels only
 
--- a/piq/fsim.py
+++ b/piq/fsim.py
@@ -147,8 +147,6 @@ def _construct_filters(x: torch.Tensor, scales: int = 4, orientations: int = 4, 
     # Pre-compute some stuff to speed up filter construction
     grid_x, grid_y = get_meshgrid((H, W), device=x.device, dtype=x.dtype)
 
-    # Move grid to GPU early on, so that all math heavy stuff computes faster.
-    # grid_x, grid_y = grid_x.to(x), grid_y.to(x)
     radius = torch.sqrt(grid_x ** 2 + grid_y ** 2)
     theta = torch.atan2(-grid_y, grid_x)
 
--- a/piq/fsim.py
+++ b/piq/fsim.py
@@ -92,8 +92,8 @@ def fsim(x: torch.Tensor, y: torch.Tensor, reduction: str = 'mean',
     pc_y = _phase_congruency(y_lum, filters=filters, scales=scales, orientations=orientations, k=k)
 
     # Gradient maps
-    filter = scharr_filter(device=x_lum.device, dtype=x_lum.dtype)
-    kernels = torch.stack([filter, filter.transpose(-1, -2)])
+    sch_filter = scharr_filter(device=x_lum.device, dtype=x_lum.dtype)
+    kernels = torch.stack([sch_filter, sch_filter.transpose(-1, -2)])
     grad_map_x = gradient_map(x_lum, kernels)
     grad_map_y = gradient_map(y_lum, kernels)
 
--- a/piq/fsim.py
+++ b/piq/fsim.py
@@ -97,7 +97,7 @@ def fsim(x: torch.Tensor, y: torch.Tensor, reduction: str = 'mean',
     )
 
     # Gradient maps
-    kernels = torch.stack([scharr_filter(), scharr_filter().transpose(-1, -2)])
+    kernels = torch.stack([scharr_filter(), scharr_filter().transpose(-1, -2)]).to(x_lum)
     grad_map_x = gradient_map(x_lum, kernels)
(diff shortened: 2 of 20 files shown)


denproc:
@snk4tr @zakajd Ready for review

piq/dss.py
```diff
@@ -162,17 +162,19 @@ def _subband_similarity(x: torch.Tensor, y: torch.Tensor, first_term: bool,
     return similarity
 
 
-def _dct_matrix(size: int) -> torch.Tensor:
+def _dct_matrix(size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r""" Computes the matrix coefficients for DCT transform using the following formula:
```
snk4tr, line 166: Please change `None` defaults to actually wanted default values. What I mean is that even though `device=None` and `dtype=None` do not break the code, they get interpreted by PyTorch as `device='cpu'` and `dtype=torch.float32` implicitly. It is not obvious that `None` casts to these values.

```suggestion
def _dct_matrix(size: int, device: str = 'cpu' dtype: type = torch.float32) -> torch.Tensor:
    r""" Computes the matrix coefficients for DCT transform using the following formula:
```

You might want to remove `from typing import Optional` as well.

piq/fsim.py
```diff
@@ -140,24 +135,33 @@ def _construct_filters(x: torch.Tensor, scales: int = 4, orientations: int = 4,
...
+    # First construct a low-pass filter that is as large as possible, yet falls
+    # away to zero at the boundaries.  All log Gabor filters are multiplied by
+    # this to ensure no extra frequencies at the 'corners' of the FFT are
+    # incorporated as this seems to upset the normalisation process when
+    # Computed explicitly without _lowpassfilter
+    # Explicit low pass filter computation
+    n = 15  # default parameter
+    cutoff = .45  # default parameter
+    assert 0 < cutoff <= 0.5, "Cutoff frequency must be between 0 and 0.5"
+    assert n > 1 and int(n) == n, "n must be an integer >= 1"
```
snk4tr, line 162: Could you elaborate why these assets are necessary considering that `cutoff` and `n` values are not parameters but hardcoded constants?

piq/functional/base.py
```diff
@@ -9,24 +9,28 @@ def ifftshift(x: torch.Tensor) -> torch.Tensor:
     return torch.roll(x, shift, tuple(range(len(shift))))
 
 
-def get_meshgrid(size: Tuple[int, int]) -> torch.Tensor:
+def get_meshgrid(size: Tuple[int, int], device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Return coordinate grid matrices centered at zero point.
```
snk4tr, line 13: Same as before: consider replacing `None` with explicit default values.

piq/functional/filters.py
```diff
@@ -1,40 +1,52 @@
 r"""Filters for gradient computation, bluring, etc."""
 import torch
 import numpy as np
+from typing import Optional
 
 
-def haar_filter(kernel_size: int) -> torch.Tensor:
+def haar_filter(kernel_size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Creates Haar kernel
```
snk4tr, line 8: Same as above.

piq/functional/filters.py
```diff
@@ -1,40 +1,52 @@
...
     """
-    kernel = torch.ones((kernel_size, kernel_size)) / kernel_size
+    kernel = torch.ones((kernel_size, kernel_size), device=device, dtype=dtype) / kernel_size
     kernel[kernel_size // 2:, :] = - kernel[kernel_size // 2:, :]
     return kernel.unsqueeze(0)
 
 
-def hann_filter(kernel_size: int) -> torch.Tensor:
+def hann_filter(kernel_size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Creates  Hann kernel
```
snk4tr, line 23: Same as above.

piq/functional/filters.py
```diff
@@ -1,40 +1,52 @@
...
-    window = torch.hann_window(kernel_size + 2, periodic=False)[1:-1]
+    window = torch.hann_window(kernel_size + 2, periodic=False, device=device, dtype=dtype)[1:-1]
     kernel = window[:, None] * window[None, :]
     # Normalize and reshape kernel
     return kernel.view(1, kernel_size, kernel_size) / kernel.sum()
 
 
-def gaussian_filter(kernel_size: int, sigma: float, dtype: torch.dtype = torch.float32) -> torch.Tensor:
+def gaussian_filter(kernel_size: int, sigma: float, device: Optional[str] = None,
+                    dtype: Optional[type] = None) -> torch.Tensor:
```
snk4tr, line 39: Same as above.

piq/functional/filters.py
```diff
@@ -45,43 +57,55 @@ def gaussian_filter(kernel_size: int, sigma: float, dtype: torch.dtype = torch.f
 
 
 # Gradient operator kernels
-def scharr_filter() -> torch.Tensor:
+def scharr_filter(device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Utility function that returns a normalized 3x3 Scharr kernel in X direction
```
snk4tr, line 61: Same as above.

piq/functional/filters.py
```diff
@@ -45,43 +57,55 @@ def gaussian_filter(kernel_size: int, sigma: float, dtype: torch.dtype = torch.f
...
     Returns:
         kernel: Tensor with shape (1, 3, 3)
     """
-    return torch.tensor([[[-3., 0., 3.], [-10., 0., 10.], [-3., 0., 3.]]]) / 16
+    return torch.tensor([[[-3., 0., 3.], [-10., 0., 10.], [-3., 0., 3.]]], device=device, dtype=dtype) / 16
 
 
-def prewitt_filter() -> torch.Tensor:
+def prewitt_filter(device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Utility function that returns a normalized 3x3 Prewitt kernel in X direction
```
snk4tr, line 73: Same as above.

piq/functional/filters.py
```diff
@@ -45,43 +57,55 @@ def gaussian_filter(kernel_size: int, sigma: float, dtype: torch.dtype = torch.f
...
+        dtype: target data type for kernel generation
     Returns:
         kernel: Tensor with shape (1, 3, 3)"""
-    return torch.tensor([[[-1., 0., 1.], [-1., 0., 1.], [-1., 0., 1.]]]) / 3
+    return torch.tensor([[[-1., 0., 1.], [-1., 0., 1.], [-1., 0., 1.]]], device=device, dtype=dtype) / 3
 
 
-def binomial_filter1d(kernel_size: int) -> torch.Tensor:
+def binomial_filter1d(kernel_size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Creates 1D normalized binomial filter
```
snk4tr, line 85: Same as above.

piq/functional/filters.py
```diff
@@ -45,43 +57,55 @@ def gaussian_filter(kernel_size: int, sigma: float, dtype: torch.dtype = torch.f
...
         Binomial kernel with shape (1, 1, kernel_size)
     """
     kernel = np.poly1d([0.5, 0.5]) ** (kernel_size - 1)
-    return torch.tensor(kernel.c).view(1, 1, kernel_size)
+    return torch.tensor(kernel.c, dtype=dtype, device=device).view(1, 1, kernel_size)
 
 
-def average_filter2d(kernel_size: int) -> torch.Tensor:
+def average_filter2d(kernel_size: int, device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
     r"""Creates 2D normalized average filter
```
snk4tr, line 99: Same as above.

piq/vsi.py
```diff
@@ -242,18 +243,20 @@ def sdsp(x: torch.Tensor, data_range: Union[int, float] = 255, omega_0: float =
     return (vs_m - min_vs_m) / (max_vs_m - min_vs_m + eps)
 
 
-def _log_gabor(size: Tuple[int, int], omega_0: float, sigma_f: float) -> torch.Tensor:
+def _log_gabor(size: Tuple[int, int], omega_0: float, sigma_f: float,
+               device: Optional[str] = None, dtype: Optional[type] = None) -> torch.Tensor:
```
snk4tr, line 247: Same as above.

tests/results_benchmark.py
```diff
@@ -356,7 +356,8 @@ def main(dataset_name: str, path: Path, metrics: List[str], batch_size: int, dev
     parser.add_argument('--metrics', nargs='+', default=[], help='Metrics to benchmark',
                         choices=list(METRICS.keys()) + list(METRIC_CATEGORIES.keys()) + ['all'])
     parser.add_argument('--batch_size', type=int, default=1, help='Batch size')
-    parser.add_argument('--device', default='cuda', choices=['cpu', 'cuda'], help='Computation device')
+    device_choice = ['cpu', 'cuda'] + [f'cuda:{i}' for i in range(torch.cuda.device_count())]
```
snk4tr, line 359: I think that we do not need to test each and every GPU because it is not our functionality. It comes from PyTorch that we rely on. Especially considering that there are no checks in this test that validate that data was sent to correct devices.

snk4tr (changes requested):
This PR practically proofs what we know from theory: direct allocation of data on correct device can save a lot of compute time. Thanks the clean job @denproc!

tests/results_benchmark.py
denproc: This change is to simplify user experience rather than testing all available devices. Old configuration doesn't allow use of non-zero device as `cuda:n` is not presented in choices. New dynamic choices helps to use a particular device for benchmarking on desired GPU.

piq/fsim.py
denproc: The hardcoded values and secretion check came from the previous implementation.
While the common use-case would be not bothering with these constants making assertions redundant, some edge cases of enthusiast wanting to vary this parameter could be tracked better.
While it is not a huge overhead in load and the same computation path as before, we do not induce additional functionality keeping an advanced user safe.

piq/dss.py
denproc: It was my initial idea to set the default `device` and `dtype` to a common `cpu` and `torch.float` explicitly. However, the standard `PyTorch` functions uses `None` as a default value. 
As we create tensors (filters or meshgrid), I believe that following the PyTorch way of creation makes more sense. In this case, the functionality of our functions `_dct_matrix` (or any filter) matches the official way `torch.zeros()` or `torch.arange()`.
If a user created a tensor with standard pytorch functions, they will expect similar behaviour from our approach.

In case of the use for library purposes, every call of the implemented functions comes with explicit `device` and `dtype` parameters.

snk4tr: I see your point @denproc. My greatest fear though is that sometime in the future PyTorch guys may decide, say, that `torch.float16` is so wide spread and efficient for ML/DL that it should become a new default. In this case, our metrics will silently break without us even knowing about it. Do you also see this possibility?

denproc: In general, this case is covered as our implementation explicitly specifies the `device` and `dtype` with every call. 

An advanced user could use `utils` functions directly for their own purposes. In this case, the change of default behaviour may hit. However, our code will still follow the PyTorch practice, i.e. delegating potential hate-wave to the big player.

snk4tr: Fair enough, merging :)

snk4tr (approved):

PR #340, opened 2023-02-06
Fix Pipelines for mypy and flake8

denproc:
Fixes `mypy` and `flake8` pipelines


snk4tr (approved):
Thank you for the quick fix! :)

PR #342, opened 2023-02-11
Optimise mesh grid for VSI

### LINKED ISSUE #335: Redundant mesh grid creation for VSI
Creation of a mesh grid for a high-volume datasets with high-res images could slow down the performance computing VSI.

In case of VSI, one mesh grid is created in `sdsp`, while another one is generated in `_log_gabor`.
As it was shown in #266 and #334, resolution of a similar issue improved the performance of the FSIM computation.

Therefore, I suggest two ideas to tackle the issue:
1. **Local**: reuse a mesh grid from `sdsp` for filter generation.
2. **Global**: mesh grids (FSIM, VSI) could be cached similar to the strategy suggested at #227.

denproc:
Closes #335 

## Proposed Changes
- cache mesh grid to be used for Gabor filter generation and generation of coordinates

### DIFF
--- a/piq/vsi.py
+++ b/piq/vsi.py
@@ -7,7 +7,7 @@ References:
 """
 import warnings
 import functools
-from typing import Union, Tuple, Optional
+from typing import Union
 
 import torch
 from torch.nn.modules.loss import _Loss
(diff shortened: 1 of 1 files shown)


snk4tr (approved):
Less redundant computations is always welcome! GJ!

PR #348, opened 2023-02-28
Metric: CLIPIQA

### LINKED ISSUE #331: Metric: CLIP-IQA and CLIP-IQA+
CLIP-IQA and CLIP-IQA+ no-reference metrics were introduced in [recent publication by Wang et.al.](https://arxiv.org/abs/2207.12396). This metric is interesting because it shows relatively good generalizability and low computational cost (for NN-based metric).

Add a class or a function with both CLIP-IQA and CLIP-IQA+ metrics.

**Additional context**
Official implementation [is available](https://github.com/IceClear/CLIP-IQA?ysclid=ld8zxmwws7970257634). However, the code has lots of dependencies and needs to be minimized.

snk4tr:
This PR implements CLIPIQA metric described in [Wang et. al. 2022](https://arxiv.org/pdf/2207.12396.pdf).
Closes #331 

The main reason to implement CLIP-IQA here is to give a user an opportunity to use the metric without bringing additional dependencies ([mmcv](https://github.com/open-mmlab/mmcv)/[mmedit](https://github.com/open-mmlab/mmediting)) to the project. 

**Note** that CLIP-IQA+ won't be implemented here because CLIP weights were fine-tuned with [mmcv](https://github.com/open-mmlab/mmcv) and hence cannot be loaded and run without it. 

**Note** that the values of this implementation correspond to the values produced by the [official CLIP-IQA implementation](https://github.com/IceClear/CLIP-IQA). SRCC scores of evaluations on public benchmarks may mismatch the ones listed in the paper. We consider official code and weights to be the ultimate source of truth and hence stick with it.

## Proposed Changes

  - Implementation of the main version of the metric (CLIP-IQA);
  - Evaluation on TID2013, KADID-10k, PIPAL, KonIQ-10k and LIVE-itW datasets (in `results_benchmark.py`);
  - Required tests. Note that values test on grayscale images are not provided because the [official single image inference script](https://github.com/IceClear/CLIP-IQA/blob/v2-3.8/demo/clipiqa_single_image_demo.py) does not support grayscale data;
  - Updates of minimal required Torchvision version in requirements and CI testing pipeline.

## Some decisions that may be questioned in the future

1. We do not use sets of prompt pairs that (as reported in the initial paper) may boost the performance of the initial CLIP-IQA metric. The reason is that the initial implementation does not use them. Hence, adding modification would mean proposing a new metric (modification of the initial one), which we try avoid here.
2. We do not allow users to provide their own prompts because of the same reason.
3. Tokens for prompt pairs are pre-computed. It allows us to avoid additional dependencies brought by tokenizer's code.
4. We had to upgrade the minimal Torchvision version because CLIP models do not work with torchvision < 0.9.1. Also, tensors computed with newer versions cannot be loaded with the old ones.


piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+        self.class_embedding = nn.Parameter(scale * torch.randn(width))
+        self.positional_embedding = nn.Parameter(scale * torch.randn((input_resolution // patch_size) ** 2 + 1, width))
+        self.ln_pre = LayerNorm(width)
+
+        self.transformer = Transformer(width, layers, heads)
+
+        self.ln_post = LayerNorm(width)
+        self.proj = nn.Parameter(scale * torch.randn(width, output_dim))
+
+    def forward(self, x: torch.Tensor, return_token=False, pos_embedding=False):
```
denproc, line 336: Should `pos_embedding` has default value `True` to match original `CLIP`? Same for `ModifiedResNet` and `AttentionPool2d`

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+        x = self.transformer(x)
+        x = x.permute(1, 0, 2)  # LND -> NLD
+        x = self.ln_final(x).type(self.dtype)
+
+        # Take features from the eot embedding (eot_token is the highest number in each sequence).
+        x = x[torch.arange(x.shape[0]), text.argmax(dim=-1)] @ self.text_projection
+
+        return x
+
+    def forward(self, image, text, pos_embedding=False, text_features=None):
```
denproc, line 506: Should `pos_embedding` has the default value equal `True` to match openai implementation, while ClipIQA could use explicit `pos_embedding=False` to discard them according to paper?

denproc (changes requested):
Great job! It's massive!
I have added some suggestions\questions in addition to code smells found by sonarcloud. 

As the PR is huge, I would suggest to wait for @zakajd review as well just to double check.

tests/test_clip_iqa.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_clip_iqa_values_rgb(clipiqa: nn.Module, device: str) -> None:
+    """Reference values are obtained by running the following script on the selected images:
+    https://github.com/IceClear/CLIP-IQA/blob/v2-3.8/demo/clipiqa_single_image_demo.py
+    """
+    paths_scores = {'tests/assets/i01_01_5.bmp': 0.45898438,
+                    'tests/assets/I01.BMP': 0.89160156}
+    for path, of_score in paths_scores.items():
+        img = Image.open(path)
+        x_rgb = torch.from_numpy(np.array(img)).float()[None]
```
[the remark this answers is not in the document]
snk4tr: Numpy haters club let's go!

tests/results_benchmark.py
```diff
@@ -9,15 +9,22 @@
...
-from typing import List, Callable, Tuple
+from typing import List, Callable, Tuple, Optional
 from pathlib import Path
 from skimage.io import imread
 from scipy.stats import spearmanr, kendalltau
 from torch.utils.data import DataLoader, Dataset
 from dataclasses import dataclass
 from torch import nn
 from itertools import chain
+from PIL import Image
```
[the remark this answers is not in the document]
snk4tr: Let's try to stick with `imread` even though I do not consider it to be important. Our benchmark script is somewhat outside of the main scope of the lib. It already requires an extensive list of dependencies and one more (really common one) wouldn't make too much of a difference.

tests/results_benchmark.py
```diff
@@ -188,14 +201,126 @@ def __init__(self, root: Path = Path("data/raw/pipal")) -> None:
...
+        transform: Something that pre-processes dataset images.
+        subset: Which part of the dataset to use.
+            Options: "train", "test", "all". "all" means train + test.
+    Returns:
+        x: image with some kind of distortion in [0, 1] range.
+        y: dummy variable, used for compatibility with FR datasets.
+        score: MOS score of the image quality.
+    """
+    _names = "AllImages_release.mat"
+    _mos = "AllMOS_release.mat"
```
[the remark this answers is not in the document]
snk4tr: I moved them to the `_filename_*` pattern but the full correspondence doesn't look feasible due to different content of the files.

tests/test_clip_iqa.py
```diff
@@ -0,0 +1,59 @@
+import torch
+import pytest
+
+import numpy as np
+
+from PIL import Image
+from piq import CLIPIQA
```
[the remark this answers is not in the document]
snk4tr: @denproc I added a test that checks that the file gets downloaded if not cached locally, which is what was not tested before. Regarding the CLIP feature extractor architecture, I lean towards not thoroughly testing it. There are two reasons for that:
1. We have a very limited usage of this architecture. For example, our metric never uses positional encoding by design. We could remove it but it is error-prone so I decided to leave it as is. All necessary blocks are tested just my running the model in existing tests.
2. The model architecture is taken from the official CLIP implementation with barely any changes. Since it is thoroughly tested by practice, I think we can be fairly confident that it works as expected.

piq/clip_iqa.py
```diff
@@ -0,0 +1,115 @@
...
+
+        Args:
+            x: An input tensor. Shape :math:`(N, C, H, W)`.
+
+        Returns:
+            The value of CLI-IQA score in [0, 1] range.
+        """
+        channels_last = x.shape[-1] in [1, 3]
+        if channels_last:
+            x = x.permute(0, 3, 1, 2)
```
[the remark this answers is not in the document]
snk4tr: @denproc I agree that we should not do explicit input tensor transformations, within any of our metrics. Regarding `_validate_input`, it does not support this functionality yet. I will add it and replace this part.

piq/clip_iqa.py
```diff
@@ -0,0 +1,113 @@
...
+This implementation is inspired by the offisial implementation but avoids using MMCV and MMEDIT libraries.
+Ref url: https://github.com/IceClear/CLIP-IQA
+"""
+import os
+import torch
+
+from torch.nn.modules.loss import _Loss
+from typing import Union
+
+from piq.feature_extractors.clip import load
```
zakajd, line 18: I would prefer importing a higher level subpackage so that it's obvious in the code below from where this `load` function came. And maybe rename it to better reflect what exactly it's loading?

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+import torch
+import torch.nn.functional as F
+
+from torch import nn
+from typing import Tuple, Union, Optional
+from collections import OrderedDict
+from urllib.request import urlopen
+
+
+CLIP_MODEL_PATH = (
```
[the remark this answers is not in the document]
zakajd: I think it's the easiest solution for now, because otherwise we'll have to use GDrive/S3 which is less convenient.

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,617 @@
...
+    "https://openaipublic.azureedge.net/clip/models/"
+    "afeb0e10f9e5a86da6080e35cf09123aca3b358a0c3e3b6c78a7b63bc04b6762/RN50.pt"
+)
+
+
+def _download(url: str, root: str) -> str:
```
zakajd, line 20: That seems to be a functionality related not only to CLIP. Can we move this to /utils? 
We already load SVM weights from one of the releases in BRISQUE code, so that can be unified?

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,617 @@
...
+
+            output.write(buffer)
+
+    if hashlib.sha256(open(download_target, "rb").read()).hexdigest() != expected_sha256:
+        raise RuntimeError("Model has been downloaded but the SHA256 checksum does not not match")
+
+    return download_target
+
+
+def load() -> nn.Module:
```
zakajd, line 60: Function name is confusing, as it's not obvious what exactly it's "loading". Given that it takes 0 params let's rename it to `load_clip_model` or similar.

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,617 @@
...
+            layers.append(Bottleneck(self._inplanes, planes))
+
+        return nn.Sequential(*layers)
+
+    def forward(self, x, return_token=False, pos_embedding=False):
+        def stem(x):
+            for conv, bn in [(self.conv1, self.bn1), (self.conv2, self.bn2), (self.conv3, self.bn3)]:
+                x = self.relu(bn(conv(x)))
+            x = self.avgpool(x)
+            return x
```
zakajd, line 236: That can be replaced by a Sequential layer initialised once during __init__. 
I don't like defining model via functions like this because of the following reasons:
1. It's harder to understand model structure 
2. It mixes model structure with model inference logic 
3. It make scripting such models harder and more error prone

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,617 @@
...
+
+        x = x.type(self.conv1.weight.dtype)
```
zakajd, line 238: Don't understand why we need this here?
In other places of the code we convert the model dtype to the input dtype, not vice versa.

zakajd (changes requested):

piq/feature_extractors/clip.py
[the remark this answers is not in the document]
denproc: I meant to suggest storing it in our repo. On one hand, it's great to download the model directly. On the other hand, if the access is closed, we wouldn't be able to use the model.
As a compromise, I would have two url addresses:
* official openaipublic
* piq repo

If the official weights are unavailable, we can download copy from our repo.

README.rst
```diff
@@ -264,13 +266,25 @@ LPIPS-VGG    0.67 / 0.67 `DISTS`_         0.72 / -                     0.57 / 0.
...
+CLIP-IQA     0.50 / -                     0.48 / -                     0.26 / -
 IS           0.26 / -                     0.25 / -                     0.09 / -
 FID          0.67 / -                     0.66 / -                     0.18 / -
 KID          0.42 / -                     0.66 / -                     0.12 / -
 MSID         0.21 / -                     0.32 / -                     0.01 / -
 GS           0.37 / -                     0.37 / -                     0.02 / -
 ===========  ===========================  ===========================  ===========================
 
+No-Reference (NR) Datasets
+^^^^^^^^^^^^^^^^^
```
denproc, line 278: This breaks the readthedocs pipeline. Length must match length of the title.
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^
```

README.rst
```diff
@@ -237,6 +237,8 @@ as it's highly dependent on fitting method and is biased towards simple examples
 
 For metrics that can take greyscale or colour images, ``c`` means chromatic version.
 
+Full-Reference (FR) Datasets
+^^^^^^^^^^^^^^^^^^^
```
denproc, line 241: This breaks the readthedocs pipeline. Length must match length of the title.
```suggestion
Full-Reference (FR) Datasets
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

README.rst
denproc, line 278: This breaks the readthedocs pipeline. Length must match length of the title.
```suggestion
No-Reference (NR) Datasets
^^^^^^^^^^^^^^^^^^^^^^^^^^
```

piq/clip_iqa.py
```diff
@@ -0,0 +1,115 @@
...
+
+        anchors = self.feature_extractor.encode_text(tokens).float()
+        self.anchors = anchors / anchors.norm(dim=-1, keepdim=True)
+
+        self.data_range = data_range
+        self.default_mean = torch.Tensor(OPENAI_CLIP_MEAN).view(1, 3, 1, 1)
+        self.default_std = torch.Tensor(OPENAI_CLIP_STD).view(1, 3, 1, 1)
+        self.logit_scale = self.feature_extractor.logit_scale.exp()
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
snk4tr: I removed the `.float()` conversion since it seems to be unnecessary (as shown above).

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+    Returns:
+        Initialized CLIP model.
+    """
+    model_path = _download(CLIP_MODEL_PATH, os.path.expanduser("~/.cache/clip"))
+
+    with open(model_path, "rb") as f:
+        model = torch.jit.load(f, map_location="cpu").eval()
+
+    model = build_model(model.state_dict())
+    model.float()
```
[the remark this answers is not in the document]
snk4tr: As discussed, leave all features of the original implementation as is.

piq/feature_extractors/clip.py
snk4tr: As discussed, leave all features of the original implementation as is.

piq/feature_extractors/clip.py
snk4tr: As discussed, leave all features of the original implementation as is.

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+def load() -> nn.Module:
+    r"""Load a CLIP model.
+
+    Returns:
+        Initialized CLIP model.
+    """
+    model_path = _download(CLIP_MODEL_PATH, os.path.expanduser("~/.cache/clip"))
+
+    with open(model_path, "rb") as f:
+        model = torch.jit.load(f, map_location="cpu").eval()
```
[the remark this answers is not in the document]
snk4tr: As discussed, leave all features of the original implementation as is.

piq/feature_extractors/clip.py
snk4tr: I am not sure I want to refactor BRISQUE code as well. It is out of scope of this PR.

piq/feature_extractors/clip.py
snk4tr: As discussed, leave all features of the original implementation as is.

snk4tr:
Ready for re-review.

denproc (commented):
Some comments for discussion

snk4tr:
For some reason I cannot comment directly reply to [this comment](https://github.com/photosynthesis-team/piq/pull/348#discussion_r1214862655) so I'll do it here. @denproc nice catch and this one actually let me guess with high probability why the original implementation had this type type conversion. It turns out that:
1. `Conv2d` only for Half precision (i.e. float16) implemented only for CUDA operations. Hence, we cannot really support it here because we want to allow both CPU and GPU computation of all our metrics.
2. There were initially two type conversions and if I remove both of them (including the one @denproc just mentioned), the abs difference in metric values per sample becomes noticeable (~1e-3). 

As a result, I think it will be fair to allow computation of the metric only in `float32` dtype. However, nothing really stops us from working on a copy of input tensor in order to not change the actual passed tensor type. I will add a comment about `float32` to the code as well.

piq/clip_iqa.py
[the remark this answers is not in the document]
snk4tr: Returned it back as described below.

piq/feature_extractors/clip.py
snk4tr: Ok, let's change the default to match the original CLIP. It might very well be my edit that I did while experimenting.

piq/utils/common.py
```diff
@@ -99,6 +101,11 @@ def _validate_input(
                 f'Expected values to be greater or equal to {data_range[0]}, got {t.min()}'
             assert t.max() <= data_range[1], \
                 f'Expected values to be lower or equal to {data_range[1]}, got {t.max()}'
+            
+        if check_for_channels_first:
+            channels_last = t.shape[-1] in {1, 2, 3}
+            assert not channels_last, "Expected tensor to have channels first format, but got channels last. \
+                Please permute channels (e.g. t.permute(0, 3, 1, 2) for 4D tensors) and rerun."
```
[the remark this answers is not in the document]
snk4tr: Hm, I agree. Now this channel order is channels first, that is what this function checks for. What exactly do you propose?

piq/clip_iqa.py
```diff
@@ -0,0 +1,117 @@
...
+        Returns:
+            The value of CLI-IQA score in [0, 1] range.
+        """
+        _validate_input(x, dim_range=(3, 4), data_range=(0., 255.), check_for_channels_first=True)
+
+        x = x.float() / self.data_range
+        x = (x - self.default_mean.to(x)) / self.default_std.to(x)
+
+        self.anchors = self.anchors.to(x)
+        self.feature_extractor = self.feature_extractor.to(x)
```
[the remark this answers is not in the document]
snk4tr: How exactly do you propose to change this particular code?

piq/utils/common.py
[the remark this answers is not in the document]
denproc: I would propose to work with `(B, C, H, W)` only and drop any functionality related to `(B, H, W, C)`. As a result, 
- we do not need to permute channels;
- we maintain consistent processing across library;
- we do not need to address different orders -> less testing.

piq/clip_iqa.py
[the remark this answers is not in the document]
snk4tr: Ok, I figured out that we can cache `anchors` and default mean/std but we can't do that to our `feature_extractor` due to the following reason: 
`TypeError: cannot assign 'piq.feature_extractors.clip.CLIP' object to buffer 'feature_extractor' (torch Tensor or None required)`

piq/feature_extractors/clip.py
```diff
@@ -0,0 +1,618 @@
...
+    def _make_layer(self, planes, blocks, stride=1):
+        layers = [Bottleneck(self._inplanes, planes, stride)]
+
+        self._inplanes = planes * Bottleneck.expansion
+        for _ in range(1, blocks):
+            layers.append(Bottleneck(self._inplanes, planes))
+
+        return nn.Sequential(*layers)
+
+    def forward(self, x, return_token=False, pos_embedding=False):
```
denproc, line 249: Another entry of `pos_embedding` to change to match [original](https://github.com/openai/CLIP/blob/a9b1bf5920416aaeaec965c25dd9e8f98c864f16/clip/model.py#L138-L154).

piq/feature_extractors/clip.py
denproc, line 523: Another entry of `pos_embedding` to change to match [original](https://github.com/openai/CLIP/blob/a9b1bf5920416aaeaec965c25dd9e8f98c864f16/clip/model.py#L358-L372).

denproc (changes requested):
I must have missed some of the entries of `pos_embedding` parameter. I think we should change it to `True` value as well to match the original clip behavioue.

piq/clip_iqa.py
```diff
@@ -0,0 +1,128 @@
...
+        # Take advantage of Torch buffers. CLIPIQA.to(device) will move these to the device as well.
+        self.register_buffer("anchors", anchors)
+        self.register_buffer("default_mean", default_mean)
+        self.register_buffer("default_std", default_std)
+
+    def forward(self, x_input: torch.Tensor) -> torch.Tensor:
+        r"""Computation of CLIP-IQA metric for a given image :math:`x`.
+
+        Args:
+            x: An input tensor. Shape :math:`(N, C, H, W)` or :math:`(C, H, W)`.
```
denproc, line 99: I would suggest to remove processing of 3D tensors to make it consistent with the rest of the metrics.

denproc:
In addition, we have to add the CLIP-IQA into the [documentation](https://github.com/photosynthesis-team/piq/blob/master/docs/source/modules.rst). This also makes me think if we have all our metrics covered in documentation. I might check it later.

piq/clip_iqa.py
snk4tr: Why would we do that it if works fine for both? I added some extra tests to verify that.

denproc (approved):
LGTM!

PR #359, opened 2023-04-21
Add meshgrid indexing

### LINKED ISSUE #358: Add indexing to pytorch meshgrid
I am receiving warnings from pytorch when using the vsi metric:
_UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument._

There is a note in the pytorch documentation for meshgrid:
_torch.meshgrid(*tensors) currently has the same behavior as calling numpy.meshgrid(*arrays, indexing=’ij’).
In the future torch.meshgrid will transition to indexing=’xy’ as the default._

It is necessary to add indexing=’ij’ in the call to meshgrid [here](https://github.com/photosynthesis-team/piq/blob/3de9a2bdd34b2da26a3e15f63b7d70a21cea9bcf/piq/functional/base.py#L34) to keep the same functionality.

rbischof:
torch.meshgrid(*tensors) currently has the same behavior as calling numpy.meshgrid(*arrays, indexing=’ij’).

In the future torch.meshgrid will transition to indexing=’xy’ as the default, so this change is necessary to keep the same functionality.

More info here: https://pytorch.org/docs/stable/generated/torch.meshgrid.html

Closes #358 

## Proposed Changes

  - Change torch.meshgrid(x, y) to torch.meshgrid(x, y, indexing='ij') [here](https://github.com/photosynthesis-team/piq/blob/3de9a2bdd34b2da26a3e15f63b7d70a21cea9bcf/piq/functional/base.py#L34)

### DIFF
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -31,7 +31,7 @@ def get_meshgrid(size: Tuple[int, int], device: Optional[str] = None, dtype: Opt
     else:
         # Even
         y = torch.arange(- size[1] / 2, size[1] / 2, device=device, dtype=dtype) / size[1]
-    return torch.meshgrid(x, y)
+    return torch.meshgrid(x, y, indexing='ij')
 
 
 def similarity_map(map_x: torch.Tensor, map_y: torch.Tensor, constant: float, alpha: float = 0.0) -> torch.Tensor:
--- a/piq/functional/base.py
+++ b/piq/functional/base.py
@@ -3,6 +3,7 @@ from typing import Tuple, Union, Optional
 import torch
 from piq.utils import _parse_version
 
+
 def ifftshift(x: torch.Tensor) -> torch.Tensor:
     r""" Similar to np.fft.ifftshift but applies to PyTorch Tensors"""
     shift = [-(ax // 2) for ax in x.size()]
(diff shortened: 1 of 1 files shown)


denproc (changes requested):
Hi @rbischof,

Thank you for raising the issue and contributing with the PR.
While the change works fine for recent and future version, it may fail for some older versions, which we still support. 

Could you please add a wrapping similar to out FFT [here](https://github.com/photosynthesis-team/piq/blob/09aad9e1bde484dfbdc4b1bc20020711145636c5/piq/vsi.py#L218) to use `indexing` for `torch>=1.10.0`?

Cheers!

piq/functional/base.py
```diff
@@ -31,7 +31,7 @@ def get_meshgrid(size: Tuple[int, int], device: Optional[str] = None, dtype: Opt
     else:
         # Even
         y = torch.arange(- size[1] / 2, size[1] / 2, device=device, dtype=dtype) / size[1]
-    return torch.meshgrid(x, y)
+    return torch.meshgrid(x, y, indexing='ij')
```
denproc, line 34:
```suggestion
    # Use indexing param depending on torch version
    recommended_torch_version = _parse_version("1.10.0")
    torch_version = _parse_version(torch.__version__)
    if len(torch_version) > 0 and torch_version >= recommended_torch_version:
        return torch.meshgrid(x, y, indexing='ij')
    return torch.meshgrid(x, y)
```

denproc (commented):
Add missing import

denproc (approved):

